PR #24, opened 2020-03-03
OP-185 | Add branching and committing guidelines to README

almeida1492:
**Resolve:** [OP-185](https://openhospital.atlassian.net/browse/OP-185)

### DIFF
--- a/README.md
+++ b/README.md
@@ -17,6 +17,63 @@ Use `npm install` to install dependencies.
 
 ## Guidelines
 
+### Branching
+
+`gitflow-tag`/`kanban-tag`-`short-description`
+
+Example: `refactor/OP-184-function-components`
+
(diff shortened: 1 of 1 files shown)


xrmx:
I think it's a bit premature to formalize these things and not sure this change would make contributing to this project easier.

mwithi:
To formalize a bit is always good, but I agree with @xrmx that at the moment we cannot perform an hard formalization for the coming contributes which would end up into refuse them.

I would suggest @almeida1492 to simplify the workflow for the "ui" to be similar to the other project's components (e.g. _<JIRA card #>-someDescription_ would be enough) and to better list in the README.md the links to the actual project's documentation we already have as Wiki in Conlfuence. In this way, the workflow is already documented there. Also, whatever is missing in Confluence could/should be added as well.

almeida1492:
I'm closing it

PR #27, opened 2020-03-22
OP-191 | Login page responsiveness and form handling

almeida1492:
Update `Login` styles in order to make it responsive. Moreover, apply `formik` to handle the form in this same activity. I hope it becomes a standard for dealing with this kind of feature from now on. 

The `SummaryPanel` could work with transitions and better positioning. It's currently static and with different screen resolutions, it could not show the desired behavior of "sliding in" when the button at the top is clicked. Also, it might show the panel content, which wasn't the idea. I didn't have time to implement that better. If someone could do that, it would be nice. 

The styles for `Login` and `SummaryPanel` are in the same file. I think it's preferable to move them into two different files. Maybe I'll do that before the merging of this Pull Request. That's why I'm marking it as WIP.d

I took this opportunity for starting the generalization of input fields with material-ui components. Both `TextFields` from the `Login` form are already implemented using a customized component that can and must be used when needed throughout the project.

I tried to line off a little bit of folder architecture under the `Login` activity and `TextField`.

**Note:** I report that the `theme` object must be configured properly. It's currently all commented on and not working properly. I'm gonna open an issue for that. 

**Resolves:** [OP-191](https://openhospital.atlassian.net/browse/OP-191?atlOrigin=eyJpIjoiMGY1YTY0MzU0MDE2NGZkZmEzNGY1N2M2ZDM2ZmI2NzUiLCJwIjoiaiJ9)


almeida1492:
@xrmx It looks like the test is failing because I renamed a file (`login.tsx `-> `Login.tsx`). Any idea of how to fix that?

alefalezza:
Honestly I think SummaryPanel could be removed from login page, we should discuss it toghether with other UX improvements.
Until then, I think it's ok to fix and improve components and global styles, theme and what else

PR #31, opened 2020-07-16
Code replacement

almeida1492:


pviotti:
Hi @almeida1492 , in the diff of this PR there are no removals of files, so if we merge it I think the repo will be a hodgepodge of both repos. I suggest you do the following: a) create a branch from the current repo b) create a commit that removes all the files c) port all the new commits to the branch by using git [cherry-pick](https://stackoverflow.com/a/5120074) or some other way to port commits.

almeida1492:
@pviotti Actually, this PR removes the files that were formerly in the master branch (note that this pull request points to `master`). If you guys think it's better, I can re-do it in a way that the new code will go to `develop` instead.

almeida1492:
Just to be sure, I opened [another PR](https://github.com/informatici/openhospital-ui/pull/32) with the code replacement targeting the `develop` branch. There you guys can see clearly that the former files were deleted and replaced with the new code.

mwithi:
Superseded by https://github.com/informatici/openhospital-ui/pull/32

PR #39, opened 2020-10-07
OP-306 | Extract sessionId from response header

almeida1492:
The way of resolving the login endpoint call response was updated in order to match the recent changes on the API YAML.


src/server.js
```diff
@@ -5,24 +5,31 @@ export function makeServer() {
...
-          credentials: {},
-          details: {},
-          name: "Marco Rossi",
-          principal: {},
-        });
+        res
+          .status(200)
+          .setHeader(
+            "set-cookie",
+            "JSESSIONID=1qrj12fcxu3a21d21pjvba6g1;Path=/oh-api"
```
alefalezza, line 19: Try adding Secure; HttpOnly to cookie

PR #54, opened 2020-10-19
OP-318 | Fix app routing

almeida1492:


axelpezzo (approved):

src/Routes.tsx
```diff
@@ -1,30 +1,35 @@
...
-import SearchPatientActivity from "./components/activities/searchPatientActivity/SearchPatientActivity";
 import PatientDetailsActivity from "./components/activities/patientDetailsActivity/PatientDetailsActivity";
+import SearchPatientActivity from "./components/activities/searchPatientActivity/SearchPatientActivity";
 
 const Routes: FunctionComponent = () => {
   return (
     <BrowserRouter basename={process.env.PUBLIC_URL}>
       <Switch>
-        <Route path="/dashboard">
+        <Route exact path="/">
```
alefalezza, line 15: I would prefer if / was assigned to dashboard, and /login to login

src/components/accessories/privateRoute/PrivateRoute.tsx
```diff
@@ -0,0 +1,23 @@
...
+const PrivateRoute: FunctionComponent<RouteProps> = ({ children, ...rest }) => {
+  const location = useLocation();
+  return (
+    <Route {...rest}>
+      {isAuthenticated() ? (
+        children
+      ) : (
+        <Redirect
+          to={{
+            pathname: "/",
```
alefalezza, line 14: This should be /login and could be a constant

alefalezza (changes requested):

src/Routes.tsx
almeida1492: I opened the task [OP-320](https://openhospital.atlassian.net/browse/OP-320) for doing this change further.

PR #85, opened 2021-02-04
Import js-resource-converter, generate resources

almeida1492:


cypress/integration/login_activity_spec.js
```diff
@@ -12,21 +12,12 @@ describe("LoginActivity spec", () => {
...
-      "Enter a valid user name"
-    );
   });
 
   it("should validate the password input", () => {
     cy.get("[id=password]").focus().clear().blur();
-    cy.get("[class=login__panel]").contains("Enter the password");
-
     cy.get("[id=password]").focus().type("123456789").blur();
-    cy.get("[class=login__panel]").should("not.contain", "Enter the password");
```
almeida1492, line 29: Considering that the language during the e2e tests is gonna be English, couldn't we keep these checks?

axelpezzo (approved):

PR #90, opened 2021-03-04
Delete npasquetto@192.168.178.15

mizzioisf:
I think this file is an output of a wrong command and should be deleted.


mwithi (approved):
Thanks!

PR #92, opened 2021-03-26
feature(patient router tabs):[OP-480] - implemented patient router tabs navigation

elvisciuffetelli:
[OP-480] - implemented patient router tabs navigation

[OP-480]: https://openhospital.atlassian.net/browse/OP-480

### DIFF
--- a/src/components/accessories/tabs/RouterTabs.tsx
+++ b/src/components/accessories/tabs/RouterTabs.tsx
@@ -1,0 +1,79 @@
+import classNames from "classnames";
+import React, { Fragment, FunctionComponent, useEffect, useState } from "react";
+import { useHistory, useRouteMatch } from "react-router-dom";
+import { Switch } from "react-router";
+import { IProps } from "./types";
+import PrivateRoute from "../privateRoute/PrivateRoute";
+import "./styles.scss";
+
+const RouterTabs: FunctionComponent<IProps> = ({ config }) => {
+  const [ currentIndex, updateCurrentIndex ] = useState(0); 
+  const history = useHistory();
+  const match = useRouteMatch();
+  const { url } = match;
+
+  const switchTab = (index: number, path: string) => {
+    updateCurrentIndex(index);
+    history.push(url + path)
+  };
+
+  const handleChangeMobileTab = (index: number): void => {
+    switchTab(index, config[index].path!)
+  };
+
+
+  useEffect(() => {
+    // default tab route - summary
+    history.push(`${url}/summary`)
+  }, [history, url]);
+
+
+  const renderHeader = (mobile = false): JSX.Element[] => {
+    if(!mobile){
+      return config.map((item, index) => {
+        const path = item.path!;
+        return (
+          <div className={classNames("tab", { active: (currentIndex === index) })} key={index} onClick={() => switchTab(index, path)}>
+            <span>{item.label}</span>
+          </div>
+        )
+      })
+    } else {
+      return config.map((item, index) => {
+        return (
+          <option className={classNames("tab", { active: (currentIndex === index) })} key={index} value={index}>
+            {item.label}
+          </option>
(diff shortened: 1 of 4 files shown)


axelpezzo (approved):

alefalezza:
We should support page refresh. Now if you refresh the page on a subroute, you are redirected to summary.
Redirection could be removed with ease (it's in a useEffect), but then you have to handle active menu item.

elvisciuffetelli:
> We should support page refresh. Now if you refresh the page on a subroute, you are redirected to summary.
> Redirection could be removed with ease (it's in a useEffect), but then you have to handle active menu item.

The problem in removing the Summary route history push from the useEffect is that we need a default one. Maybe a solution can be to save the current route in the session storage and then recover it from there after reloading the page?

alefalezza:
> > We should support page refresh. Now if you refresh the page on a subroute, you are redirected to summary.
> > Redirection could be removed with ease (it's in a useEffect), but then you have to handle active menu item.
> 
> The problem in removing the Summary route history push from the useEffect is that we need a default one. Maybe a solution can be to save the current route in the session storage and then recover it from there after reloading the page?

I think it would be better to use current location to select active tab, it also supports deep linking, which is nice.
I think "currentIndex" could be string (`url + item.path`) instead of number, and current tab should match current location.

Even better, match could be evaluated with a regexp like `/^(location_here)([\/?].*)?$/`, which matches:
- "location"
-  "location/"
-  "location/sub"
-  "location?query" 

but not:
- "location/sub1"
- "location/sub1/"
- "location/sub1?query"

alefalezza:
> > > We should support page refresh. Now if you refresh the page on a subroute, you are redirected to summary.
> > > Redirection could be removed with ease (it's in a useEffect), but then you have to handle active menu item.
> > 
> > 
> > The problem in removing the Summary route history push from the useEffect is that we need a default one. Maybe a solution can be to save the current route in the session storage and then recover it from there after reloading the page?
> 
> I think it would be better to use current location to select active tab, it also supports deep linking, which is nice.
> I think "currentIndex" could be string (`url + item.path`) instead of number, and current tab should match current location.
> 
> Even better, match could be evaluated with a regexp like `/^(location_here)([\/?].*)?$/`, which matches:
> 
> * "location"
> * "location/"
> * "location/sub"
> * "location?query"
> 
> but not:
> 
> * "location/sub1"
> * "location/sub1/"
> * "location/sub1?query"

I think current path could be evaluated as:

```typescript
const {pathname} = useLocation();
const currentPath: string | undefined = config.map(item => item.path).find(path => pathname.match(new RegExp(`^(${url}${path})([/?].*)?$`, "gi")));
```

elvisciuffetelli:
@alefalezza I understand regarding the active tab. But what about the first route pushed to the history when we land on the page? If we remove this useEffect:
` useEffect(() => {
    // default tab route - summary
    history.push(`${url}/summary`)
  }, [history, url]);`

there won't be any tab route pushed to the history, so no tab will be rendered until the user click on a tab

alefalezza:
> @alefalezza I understand regarding the active tab. But what about the first route pushed to the history when we land on the page? If we remove this useEffect:
> 
> ```js
>  useEffect(() => {
>     // default tab route - summary
>     history.push(`${url}/summary`)
>   }, [history, url]);
> ```
> 
> there won't be any tab route pushed to the history, so no tab will be rendered until the user click on a tab

I suggest you:

- to parametrize `/summary` using an optional prop, so it can be injected and RouterTabs reusability increases
- to handle default route adding a check on currentRoute, and going to default only if current tab is undefined AND default route is defined (that's why it's an optional prop)

elvisciuffetelli:
Ok got it, now it supports page refresh  👍

alefalezza:
Perfect, thank you @elvisciuffetelli !

PR #93, opened 2021-04-02
Feature/op 363 setup permissions

alefalezza:
I imported new spec with permissions (/users/me api) and launch generator. 
Then I started implementing permission logic: it's a working purpose, feel free to suggest changes.

Permissions are described in a typescript definition, even though they are dynamic, but we need to control them statically.

A PermissionProvider (not really a react provider) is a component responsible to inizialize permissions using redux. It reloads when authenticate user changes. It could also save permissions to localStorage, for better performances.
It should handle loading state and errors.

Then, a usePermission Hook is used to check permissions in business components.

A temporary PermissionDenied component handles visualization (and eventually logic) for unauthenticated accesses.

I also added default prettier configuration, to activate VSCode default formatter.


src/libraries/permissionUtils/usePermission.ts
```diff
@@ -0,0 +1,10 @@
+import { useSelector } from "react-redux";
+import { IState, TPermission } from "../../types";
+
+export const usePermission = (name: TPermission): boolean => {
+  const permissions = useSelector<IState, string[]>(
+    (state) => state.main.me?.data?.permission || []
+  );
+
+  return Boolean(permissions.find((permission: string) => permission === name));
```
almeida1492, line 9: Why do we have to use a `Boolean` object here? Maybe a `!!` would do the job in a more performant fashion, wouldn't it?
```suggestion
  return !!permissions.find((permission: string) => permission === name);
```

almeida1492 (commented):
It looks neat. The only thing I'd change here is the point of implementation of the `usePermission` hook. I mean, in your proposal, you apply it inside the component that requires some permission in order to be rendered. I would use it in a layer above that component, to hide or not the button that would take the user to that activity. By doing so, we don't need to display feedback of permission denied to the user.

src/libraries/permissionUtils/usePermission.ts
alefalezza: I don't know what is more performant, as long as there is a performance problem...

src/libraries/permissionUtils/PermissionProvider.tsx
```diff
@@ -0,0 +1,24 @@
...
+  const authenticatedUser = useSelector<IState, string | undefined>(
+    (state) => state.main.authentication?.data?.displayName
+  );
+
+  useEffect(() => {
+    if (!authenticatedUser) {
+      return;
+    }
+    // reload permissions when user authenticates
+    getUserInfoThunk()(dispatch);
```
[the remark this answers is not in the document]
alefalezza: It was a POC, I think this hook could be useful to render "edit" and "delete" buttons in the future, so I would keep it as an option.  

For this purpose it would be better to wrap components in a component (or HOC?) which takes care of conditional rendering and fallback. I think this check should be preserved to control deep linking.

I agree with you that an additional control should be applied to menu (or tabs).

src/components/activities/patientDetailsActivity/tabsConfig.tsx
```diff
@@ -4,13 +4,48 @@ import SkeletonLoader from "../../accessories/skeletonLoader/SkeletonLoader";
...
-  { label: "Triage", path:"/triage", content: <PatientDetailsContent title="Triage" content={<PatientTriage />} /> },
-  { label: "Therapy", path:"/therapy", content: <PatientDetailsContent title="Therapy" content={<SkeletonLoader />} /> },
-  { label: "Booking", path:"/booking", content: <PatientDetailsContent title="Booking" content={<SkeletonLoader />} /> },
+  {
+    label: "Summary",
+    path: "/summary",
+    content: (
+      <PatientDetailsContent
+        title="Summary"
+        content={withPermission("summary.read")(PatientSummary)}
```
alefalezza, line 17: @almeida1492 I tried to pull up this check, but going further would mean:
- to change TTabConfig signature (it's ok for me)
- to add a loader indicator in permission provider, because permissions are empty while loading, and current logic of this component would redirect to default route

alefalezza:
- remove "PermissionPrivoder" and couple authentication and permission logic, now handled by a single "setAuthenticationThunk" action and a single reducer case
- add permissions to SessionStorage, and provide functions to rehydrate auth state from session
- extract redirection after authentication in a component "RedirectAfterLogin" (more explicit)
- add function "getAuthenticationFromSession" to avoid code duplication
- centralize session restore in "useAuthentication" hook. It has been reinforced by aligning redux and session data

PR #153, opened 2021-06-03
Api logic/summary op 531 rxjs

tsognong:


src/components/accessories/patientSummary/patientSummaryByDate/PatientSummaryByDate.tsx
```diff
@@ -1,43 +1,72 @@
...
+  hasSucceeded,
+  loadSummaryData,
+  isLoading,
+}) => {
+  const [summaryData, setSummaryData] = useState(Array<SummaryData>());
+  const summary = useSelector<IState, ISummaryState>((state) => state.summary);
+  const patient = useSelector((state: IState) => state.patients);
+
+  useEffect(() => {
+    loadSummaryData(patient.selectedPatient?.data?.code!);
```
almeida1492, line 28: Here we should handle the case in which the code is undefined, instead of using a non-null assertion like that.

src/components/accessories/patientSummary/patientSummaryByDate/PatientSummaryByDate.tsx
```diff
@@ -1,43 +1,72 @@
...
+  }, [patient]);
 
-const PatientSummaryByDate: FunctionComponent = () => {
+  useEffect(() => {
+    if (hasSucceeded) {
+      setSummaryData(summary.loadSummaryData.data!);
+    }
+  }, [summary]);
```
almeida1492, line 35: It seems that this useState (I mean, the one in line 23) is unnecessary because we can handle all this logic regarding summaryData inside redux boundaries. This component, then, would see only a useSelector that grabs the portion of the store that we need.

src/components/accessories/patientSummary/patientSummaryByDate/PatientSummaryByDate.tsx
```diff
@@ -1,43 +1,72 @@
...
+          />
+        )}
+      </div>
+    </>
   );
-}
+};
+
+const mapStateToProps = (state: IState): IStateProps => ({
+  isLoading: state.summary.loadSummaryData.status === "LOADING",
```
almeida1492, line 60: We should try to use verbs for naming variables only if those are function references. In this case, it's an object, so we could call it summaryData or something like that.

almeida1492 (changes requested):

src/components/accessories/patientSummary/patientSummaryByType/PatientSummaryByType.tsx
```diff
@@ -1,94 +1,125 @@
...
-const PatientSummaryByType: FunctionComponent = () => {
+const PatientSummaryByType: FunctionComponent<TProps> = ({
+  hasSucceeded,
+  loadSummaryData,
+  isLoading,
+}) => {
+  const { t } = useTranslation();
+  const [summaryData, setSummaryData] = useState(Array<SummaryData>());
+  const summary = useSelector<IState, ISummaryState>((state) => state.summary);
+  const patient = useSelector((state: IState) => state.patients);
```
almeida1492, line 31: The same comment in PatientSummaryByDate goes here. We need to avoid this overhead.

src/components/accessories/patientTherapy/PatientTherapy.tsx
```diff
@@ -55,7 +55,7 @@ const PatientTherapy: FC<TProps> = ({
 
   const onSubmit = (therapy: TherapyRowDTO) => {
     setShouldResetForm(false);
-    therapy.patID = undefined;
+    therapy.patID = patient;
```
almeida1492, line 58: Maybe it would be better to name `patient` as `patientId`, in order to make it clearer.

src/libraries/reduxUtils/convert.ts
```diff
@@ -0,0 +1,15 @@
+import { format } from "date-fns";
+import { SummaryFieldType } from "./SummaryFieldType";
+
+export const convertToSummaryData = (
```
almeida1492, line 4: We could name the file with the very same name of the function inside it.

src/mockServer/fixtures/diseaseDTO.js
```diff
@@ -0,0 +1,13 @@
+const disease1DTO = {
```
almeida1492, line 1: Maybe there's a typo in this identifier?

src/resources/i18n/am.json
```diff
@@ -116,5 +116,34 @@
     "patientregistrationsuccessfull": "The patient registration was successful.",
     "somethingwrong": "Something went wrong, please retry later.",
     "searchpatientinstruction": "If you don't have the patient ID, fill as many fields as possible and click search"
+  },
```
almeida1492, line 119: There's no need to repeat these properties in every i18n resource file. The application uses English by default, so you just need to add these new strings to `en.json`.

almeida1492 (changes requested):

src/components/accessories/patientSummary/patientSummaryByDate/PatientSummaryByDate.tsx
tsognong: Ok, right.

src/resources/i18n/am.json
tsognong: I understand.

src/components/accessories/patientSummary/patientSummaryByDate/PatientSummaryByDate.tsx
tsognong: Ok, I replace that by: 
` if (patient.selectedPatient.data?.code)
      loadSummaryData(patient.selectedPatient.data.code);`

src/components/accessories/patientSummary/patientSummaryByType/PatientSummaryByType.tsx
tsognong: OK, I applied it. Thanks

src/components/accessories/patientSummary/patientSummaryByDate/PatientSummaryByDate.tsx
tsognong: It is OK

src/components/accessories/patientSummary/patientSummaryByDate/PatientSummaryByDate.tsx
tsognong: I replaced it.

src/components/accessories/patientTherapy/PatientTherapy.tsx
tsognong: I replaced with patientId, but the confusion there comes from therapy api data model. Take a look: 
`
{
  "therapyID": 1,
  "patID": {
    "code": 0,
    .........
    }
    .........
}
`

tsognong: So the naming in api data model is not fine.

src/libraries/reduxUtils/convert.ts
tsognong: Ok, in that case the file just belongs to that function. I name like that because we could add another function inside.

src/mockServer/fixtures/diseaseDTO.js
tsognong: I replaced that to `
 export const diseaseDTO = {....}
`
but the intention there was to create several objects, desease1, desease2, desease3 for mock purpose

src/components/accessories/patientSummary/patientSummaryByDate/PatientSummaryByDate.tsx
```diff
@@ -16,24 +16,21 @@ const label = {
...
-  hasSucceeded,
   loadSummaryData,
   isLoading,
 }) => {
-  const [summaryData, setSummaryData] = useState(Array<SummaryData>());
-  const summary = useSelector<IState, ISummaryState>((state) => state.summary);
+  const summaryData = useSelector<IState, Array<SummaryDataType>>((state) =>
+    state.summary.summaryData.data
+      ? state.summary.summaryData.data
+      : new Array<SummaryDataType>()
```
almeida1492, line 25: There's no need to do that. You can declare a value for `summary` in the redux initial state. Also, there's some naming confusion here. The structure should be something like `state.summary.data` and the variable inside the component `PatientSummaryByDate` could simply be declared under the identifier `summary`.

almeida1492 (changes requested):

src/components/accessories/patientSummary/patientSummaryByDate/PatientSummaryByDate.tsx
tsognong: Ok, I improve it.

tsognong: I remove that variable declaration there.

almeida1492 (approved):

PR #156, opened 2021-06-10
Api logic/op 534 therapy

tsognong:


src/components/accessories/patientTherapy/PatientTherapy.tsx
```diff
@@ -24,18 +26,22 @@ import checkIcon from "../../../assets/check-icon.png";
...
   const { t } = useTranslation();
   const infoBoxRef = useRef<HTMLDivElement>(null);
   const [shouldResetForm, setShouldResetForm] = useState(false);
   const [shouldUpdateTable, setShouldUpdateTable] = useState(false);
   const [activityTransitionState, setActivityTransitionState] =
     useState<TherapyTransitionState>("IDLE");
-  const patient = useSelector(
+
+  //There is some "mistake" in apis model about the patient id
+  const patientId = useSelector(
```
axelpezzo, line 44: If the patientId variable contains the patient data, then rename it in patientData in order to improve the semanticity.

axelpezzo: Or extract only the patient.code and leave the name as before.

src/components/accessories/patientTherapy/patientTherapyTable/PatientTherapyTable.tsx
```diff
@@ -13,24 +11,25 @@ const PatientTherapyTable: FunctionComponent<IOwnProps> = ({
...
+    if (newTherapyObj) {
+      const found = data.find(
+        (item) => item.medicalId == newTherapyObj?.medicalId
+      );
+      if (!found) {
+        data = [...data, newTherapyObj];
+      }
+    }
+  };
+  useSelector<IState>((state) => updateTableData(state));
```
axelpezzo, line 32: I would put the data variable inside a useState and I'd delete the setUpdate method.

So, when the shouldUpdateTable will be updated, inside the useEffect you'll perform the Redux call updating the data with an setData.

This should be more readable, what do you think?

src/components/accessories/patientTherapy/therapyForm/TherapyForm.tsx
```diff
@@ -191,7 +200,7 @@ const TherapyForm: FC<TherapyProps> = ({
               <DateField
                 fieldName="startDate"
                 fieldValue={formik.values.startDate}
-                disableFuture={true}
+                disableFuture={false}
```
axelpezzo, line 203: Is this assertion right? Don't we need the future date?

src/components/accessories/patientTherapy/therapyForm/styles.scss
```diff
@@ -1,117 +1,74 @@
 @import "../../../../styles/variables";
 @import "../../../../../node_modules/susy/sass/susy";
-
+$ipadpro: 1024px 1366px;
```
axelpezzo, line 3: This isn't the right place to initialize the variable. Anyway you shouldn't use this but $narrow instead. It should be cover from 1025px to 1200px if I'm right.

src/libraries/formDataHandling/functions.ts
```diff
@@ -14,7 +14,7 @@ export const getFromFields = (
 
 const parseDate = (raw: string) => {
   const unformatDate = new Date(+raw).toString();
-  return Date.parse(unformatDate).toString();
+  return raw === "" ? "" : Date.parse(unformatDate).toString();
```
axelpezzo, line 17: `return raw ? Date.parse(unformatDate).toString() : "";`

More readable

src/state/therapies/actions.ts
```diff
@@ -32,13 +33,17 @@ export const createTherapy =
       (payload) => {
         dispatch({
           type: CREATE_THERAPY_SUCCESS,
-          payload,
+          payload: {
+            ...payload,
+            startDate: format(new Date(+payload.startDate!), "dd/MM/yyyy"),
+            endDate: format(new Date(+payload.endDate!), "dd/MM/yyyy"),
```
axelpezzo, line 39: I think that should be more practical check if the payload.startDate or payload.endDate both exists before convert them into a Date.

src/state/therapies/reducer.ts
```diff
@@ -25,6 +25,10 @@ export default produce((draft: ITherapiesState, action: IAction<any, any>) => {
     case CREATE_THERAPY_SUCCESS: {
       draft.createTherapy.status = "SUCCESS";
       draft.createTherapy.data = action.payload;
+      draft.therapiesByPatientId.data = [
+        ...draft.therapiesByPatientId.data!,
+        action.payload,
+      ];
```
axelpezzo, line 31: The ! operator could cause some uncontrolled effects. It should more practical here to handler directly the case when the data will be null.

src/state/therapies/actions.ts
axelpezzo: Maybe I miss the point but when or where are you gonna use startDate and endDate?

src/components/accessories/patientTherapy/PatientTherapy.tsx
tsognong: It is right. But I will still have this syntaxe

`const onSubmit = (therapy: TherapyRowDTO) => {
    setShouldResetForm(false);
    therapy.patID = patientData; // there is some mistake in api data model naming
    createTherapy(therapy);
  };`

because of some mistake in endpoint data model.

src/components/accessories/patientTherapy/patientTherapyTable/PatientTherapyTable.tsx
tsognong: It is right.

src/components/accessories/patientTherapy/therapyForm/TherapyForm.tsx
tsognong: Here I use the false value to display future date. When it is true the future dates don't display.

src/components/accessories/patientTherapy/therapyForm/styles.scss
tsognong: I will remove it, because the $narrow variable is already there. But the responsive is still unresolved

src/libraries/formDataHandling/functions.ts
tsognong: ok.

src/state/therapies/actions.ts
tsognong: The created data will be used to update the therapy table, instead of refreshing the whole table each time a new data is created.

src/state/therapies/reducer.ts
tsognong: You are right. I will replace it by 
`
 if (draft.therapiesByPatientId.data)
        draft.therapiesByPatientId.data = [
          ...draft.therapiesByPatientId.data,
          action.payload,
        ];
`

src/components/accessories/patientTherapy/patientTherapyTable/PatientTherapyTable.tsx
tsognong: It is not possible to use useSelector inside useEffect

src/components/accessories/patientTherapy/PatientTherapy.tsx
axelpezzo: It's ok as long as we won't fix the APIs.

src/components/accessories/patientTherapy/patientTherapyTable/PatientTherapyTable.tsx
axelpezzo: Yes, you're right. But, if I'm not wrong, you can access to the state outside the useEffect, maybe setting a variable state. Am I right or I'm just overthinking?

src/components/accessories/patientTherapy/therapyForm/styles.scss
axelpezzo: I already fixed that inside another branch which it's not merged yet. The fault was to the Table under the form and not the form itself.

src/state/therapies/reducer.ts
axelpezzo: If the condition is false, it's OK if I will not pass anything to the draft.therapiesByPatientId.data? Won't this cause any unhandled behaviours?

src/components/accessories/patientTherapy/patientTherapyTable/PatientTherapyTable.tsx
```diff
@@ -1,35 +1,37 @@
...
+};
+export const order = ["startDate"];
 
 const PatientTherapyTable: FunctionComponent<IOwnProps> = ({
   shouldUpdateTable,
 }) => {
-  const [, setUpdate] = useState(false);
-  const therapy = useSelector<IState, ITherapiesState>(
-    (state) => state.therapies
+  let therapyStore = useSelector<IState, TherapyRowDTO[]>((state) =>
```
axelpezzo, line 25: Here you can use const instead of let, after that should be OK.

src/state/therapies/reducer.ts
```diff
@@ -25,6 +25,12 @@ export default produce((draft: ITherapiesState, action: IAction<any, any>) => {
     case CREATE_THERAPY_SUCCESS: {
       draft.createTherapy.status = "SUCCESS";
       draft.createTherapy.data = action.payload;
+      if (draft.therapiesByPatientId.data)
+        draft.therapiesByPatientId.data = [
+          ...draft.therapiesByPatientId.data,
+          action.payload,
+        ];
+      else draft.therapiesByPatientId.data = [action.payload];
```
axelpezzo, line 33: Please, here add the curly brackets to the if / else statement. After that shloud be OK.

src/libraries/sortUtils/sortUtils.ts
```diff
@@ -1,17 +1,22 @@
...
-      moment(a[orderBy]).format("YYYY-MM-DD")
-      ? 1
-      : -1;
+const formats = [
+  "DD/MM/YYYY",
+  "YYYY-MM-DD",
+  "YYYY-MM-DD HH:mm:ss",
+  "DD/MM/YYYY HH:mm:ss",
+];
+export function DateComparator<T>(a: T, b: T, orderBy: keyof T) {
```
almeida1492: Functions should be named as verbs and shouldn't be capitalized.

src/state/therapies/reducer.ts
tsognong: I added the if() and else,  thanks.

PR #157, opened 2021-06-16
Feature/opd report op 551

tsognong:


src/components/accessories/patientOPD/patientOPDTable/PatientOPDTable.tsx
```diff
@@ -0,0 +1,68 @@
...
+
+  const onEdit = () => {
+    console.log("update");
+  };
+
+  const onEView = () => {};
+
+  useEffect(() => {
+    setUpdate(shouldUpdateTable);
+  }, [shouldUpdateTable]);
```
axelpezzo, line 47: Here the same as the previous Merge Request. Put the data inside an useState statement.

src/components/accessories/table/Table.tsx
```diff
@@ -28,6 +25,7 @@ const Table: FunctionComponent<IProps> = ({
   isCollapsabile,
   rowsPerPage,
   columnsOrder,
+  compareRowBy = defaultComparator,
```
almeida1492, line 28: Maybe we could call it "compareRows" or something like that

almeida1492 (changes requested):

PR #162, opened 2021-07-16
fix fields validation + options translation OP-548

tsognong:


src/components/accessories/patientDataForm/PatientDataForm.tsx
```diff
@@ -37,10 +37,14 @@ const PatientDataForm: FunctionComponent<TProps> = ({
...
-    secondName: string().required("This field is required"),
-    birthDate: string().required("This field is required"),
-    sex: string().required("This field is required"),
+    firstName: string().required(t("common.required")),
+    secondName: string().required(t("common.required")),
+    birthDate: string().required(t("common.required")),
+    sex: string().required(t("common.required")),
+    telephone: string().matches(
+      /^[\+]?[(]?[0-9]{3}[)]?[-\s\.]?[0-9]{3}[-\s\.]?[0-9]{4,6}$/,
+      t("common.incorrectformat")
```
axelpezzo, line 46: It seems fine but as I said this morning, we must test all the different number format.
If it doesn't,we could simple add a Regex that check if the charaters inserted are only numbers.

**Suggestion:**
Maybe we could add some extra information below the fields as for example: info about the accepted phone numbers formats.

src/components/accessories/selectField/SelectField.tsx
```diff
@@ -50,7 +51,7 @@ const SelectField: FunctionComponent<IProps> = ({
         </MenuItem>
         {options.map((option, index) => (
           <MenuItem value={option.value} key={index}>
-            {option.label}
+            {t(option.label)}
```
axelpezzo, line 54: We are not gonna translate locally the api's value. We must communicate with the Back end and pass in some way the app's language.

tsognong: There are some values that we have to translate in ui like Y/N options, male/femele options... I think that we can keep this code at the moment.

axelpezzo: I understand the purpose behind this but we could improve the code in any case: I don't like that if we're not gonna use the translate function we invoke the function in any case, this could lead to some unexpected behiovurs. Maybe we could optimize this in some way?

axelpezzo (approved):

PR #164, opened 2021-07-22
begin auto complete feature

tsognong:


src/components/accessories/autocompleteField/AutocompleteField.tsx
```diff
@@ -0,0 +1,65 @@
...
+      <Autocomplete
+        freeSolo
+        options={options}
+        getOptionLabel={(option) => option.label}
+        value={geFullObj(value)}
+        getOptionSelected={(option, v) => option.value === v.value}
+        renderOption={(option) => <Fragment>{option.label}</Fragment>}
+        onChange={(e: object, val: any | null) => {
+          setValue(val?.value || "");
+        }}
```
axelpezzo, line 47: We should add a debounce function in order to prevent the status updated every single time a user types on a key.

Something like that:

`onChange={(e: object, val: any | null) => {
  debounceUpdate(val?.value || "");
}}

const debounceUpdate= useCallback(
  debounce(
    (value) =>
      setValue(value),
    250
  ),
  []
);
`

axelpezzo: I'm not sure if the useCallback couldn't cause some throubles, if it does, try to remove it.

src/components/accessories/dateField/DateField.tsx
```diff
@@ -25,7 +26,13 @@ const DateField: FunctionComponent<IProps> = ({
...
 
   useEffect(() => {
-    setValue(fieldValue === "" ? null : new Date(+fieldValue));
+    if (fieldValue === "") {
+      setValue(null);
+    } else if (moment(+fieldValue).isValid()) {
+      setValue(moment(+fieldValue).toDate());
+    } else {
+      setValue(moment(fieldValue, "DD/MM/YYYY").toDate());
+    }
```
axelpezzo, line 35: See the comment on #165 MR

axelpezzo (approved):

PR #165, opened 2021-07-26
Feature/op 585 edit opd

tsognong:


src/components/accessories/dateField/DateField.tsx
```diff
@@ -25,7 +26,15 @@ const DateField: FunctionComponent<IProps> = ({
...
-    setValue(fieldValue === "" ? null : new Date(+fieldValue));
+    if (fieldValue === "") {
+      setValue(null);
+    } else if (moment(+fieldValue).isValid()) {
+      setValue(moment(+fieldValue).toDate());
+    } else if (moment(fieldValue, "DD/MM/YYYY").isValid()) {
+      setValue(moment(fieldValue, "DD/MM/YYYY").toDate());
+    } else {
+      setValue(moment(fieldValue).toDate());
+    }
```
axelpezzo, line 37: Here man sorry but I can't understand what is the purpose of this.

src/components/accessories/patientOPD/patientOPD.tsx
```diff
@@ -1,137 +1,158 @@
...
-          />
+        {createStatus === "FAIL" && (
+          <InfoBox type="error" message={t("common.somethingwrong")} />
         )}
       </div>
+      <div ref={infoBoxRef}>
+        {updateStatus === "FAIL" && (
+          <InfoBox type="error" message={t("common.somethingwrong")} />
+        )}
+      </div>
```
axelpezzo, line 131: `<div ref={infoBoxRef}>` should be inside the if statement.

src/resources/i18n/en.json
```diff
@@ -45,7 +45,14 @@
...
     "disease3": "Diagnostic 3",
-    "validatedisease": "This diagnostic should be different"
+    "validatedisease": "This diagnostic should be different",
+    "code": "Code",
+    "updateopd": "update opd",
+    "saveopd": "save opd",
+    "created": "Opd Created",
+    "updated": "Opd updated",
+    "createsuccess": "Opd has been created successfully!",
+    "updatesuccess": "Opd {{code}} has been updated successfully!"
```
axelpezzo, line 55: Here I wouldn't use the code, is there something else of more useful inside the object?

src/components/accessories/dateField/DateField.tsx
tsognong: The purpose is to fill  the date field with the fieldValue prop. The problem here is the format. This old code `
setValue(fieldValue === "" ? null : new Date(+fieldValue));` is not working for edit case. For example when you pass "30/08/1991" in edit patient mock data, it does'nt work. So we have to convert data fieldValue before setting.

src/components/accessories/patientOPD/patientOPD.tsx
tsognong: Ok, I apply it right now.

src/resources/i18n/en.json
tsognong: May be the date? But the date is not unique. May be the diseases names? It could take more place than expected. May be the complain? It could take more place and can also be empty. That's why I choose the code property.

src/components/accessories/dateField/DateField.tsx
axelpezzo: The mocked value should be the same as the patientDTO, the date should comes from the server as a timestamp string.

The dateField component must be agnostic than the date format used.

src/resources/i18n/en.json
axelpezzo: Ok we definitely keep the code. If we didn't do yet so in this case we should put the code inside the OPD Table in order to avoid some users delete mistake.

tsognong: I already put the code in the table.

src/components/accessories/dateField/DateField.tsx
tsognong: OK. the date used in cypress mock for edit patient was   "birthDate": "30/08/1991", So I will also replace that to timestamp.

tsognong: I think at least we have to consider two cases for date: timestamp and millisecondes date. So I update the dates converters according to that two possibilities.

src/components/accessories/dateField/DateField.tsx
```diff
@@ -25,7 +26,13 @@ const DateField: FunctionComponent<IProps> = ({
...
 
   useEffect(() => {
-    setValue(fieldValue === "" ? null : new Date(+fieldValue));
+    if (fieldValue === "") {
+      setValue(null);
+    } else if (moment(+fieldValue).isValid()) {
+      setValue(moment(+fieldValue).toDate());
+    } else {
+      setValue(moment(fieldValue).toDate());
+    }
```
axelpezzo, line 35: I continue to not be sure about this. I tested the edit patient locally and it seems like to work as expected with the previous code. Maybe we could restore the previous code and then we could test again which is the problem you reported.

Also, the the if statement it's not properly formatted: you run moment.toDate even if the date is not valid?

tsognong: @axelpezzo Could you please provide the tag of the commit concerning your working code? Because before working on patient form validation, I tested the edit feature and the birthday did'nt get value. So on what commit tag did your code working as exepected? 

For the else statement I agree with you that I should check the date validity.

axelpezzo: I mean, the previous code I wrote should be correct. The issue here is the date format that comes from the API that is not in millisecond but it's a date string formatted. The issue should be discussed with the back-enders in order to standardize the date format in all his occurency in something of reusable and consistent.

axelpezzo: The date should always be a timestamp and not a string format date as we already done inside the mockServer API in patientDTO spec.

tsognong: May be you could discuss with Henrique, so as soon as possible I apply what you decide together. Because it is at the moment the only point that is bloking my PR here.

axelpezzo: I talked with Henrique about this and he agrees with me about this: 

- the API should return only one date format and it should be timestamp averywhere.
- if, the point one it's not possible, this convertion must be outside this componente as for example when you process the date from the API. This component should receive a timestamp in any case.

So, i'm aking to you to restore the previous code and eventually, update the mockData of the opd response with a timestamp as we already done for the patientDTO spec.

tsognong: OK.

axelpezzo (approved):

axelpezzo:
@tsognong Same here, solve the conflits and then I'll merge the branch in develop.

PR #166, opened 2021-07-29
Feature/op 586 edit therapy

tsognong:


src/components/accessories/dateField/DateField.tsx
```diff
@@ -25,7 +26,13 @@ const DateField: FunctionComponent<IProps> = ({
...
 
   useEffect(() => {
-    setValue(fieldValue === "" ? null : new Date(+fieldValue));
+    if (fieldValue === "") {
+      setValue(null);
+    } else if (moment(+fieldValue).isValid()) {
+      setValue(moment(+fieldValue).toDate());
+    } else {
+      setValue(moment(fieldValue).toDate());
+    }
```
axelpezzo, line 35: See the comment on #165 MR

src/libraries/formatUtils/dataFormatting.ts
```diff
@@ -14,9 +15,15 @@ export const opdDataFormatter = (
     diseases && data.disease3 !== ""
       ? diseases.find((el) => el.code === +data.disease3)
       : null;
```
axelpezzo, line 17: There is a reason why are you using a unary plus operator so frequently inside this function?

src/libraries/formatUtils/dataFormatting.ts
```diff
@@ -14,9 +15,15 @@ export const opdDataFormatter = (
...
-
   return data;
 };
+
+export const renderDate = (date: string) => {
+  return moment(date).isValid()
+    ? moment(date).format("DD/MM/YYYY")
+    : moment(+date).isValid()
+    ? moment(+date).format("DD/MM/YYYY")
+    : "";
```
axelpezzo, line 28: I think this function could be write in this way:  

```
return moment(date).isValid()
    ? moment(date).format("DD/MM/YYYY")
    : "";
```

Do you agree?

src/libraries/reduxUtils/convert.ts
```diff
@@ -7,7 +7,9 @@ export const convertToSummaryData = (
 ) => {
   const res = data.map(({ ...rest }) => ({
     type: field.type,
-    date: moment(rest[field.dateField]).format("DD/MM/YYYY"),
+    date: moment(rest[field.dateField]).isValid()
+      ? moment(rest[field.dateField]).format("DD/MM/YYYY")
+      : moment(+rest[field.dateField]).format("DD/MM/YYYY"),
```
axelpezzo, line 12: Are you trying to format a date even if it's invalid?

src/components/accessories/patientTherapy/therapyForm/TherapyForm.tsx
```diff
@@ -41,7 +42,10 @@ const TherapyForm: FC<TherapyProps> = ({
         name: "endDate",
         message: t("therapy.validatelastdate"),
         test: function (value) {
-          return moment(value).isSameOrAfter(moment(this.parent.startDate));
+          if (moment(+value).isValid()) {
+            return moment(+value).isSameOrAfter(moment(+this.parent.startDate));
+          } else
+            return moment(value).isSameOrAfter(moment(this.parent.startDate));
```
axelpezzo, line 48: As below, if the date is invalid it shouldn't be formatted.

src/components/accessories/patientTherapy/therapyForm/TherapyForm.tsx
```diff
@@ -1,4 +1,5 @@
 import { Checkbox, FormControlLabel } from "@material-ui/core";
+import id from "date-fns/esm/locale/id/index.js";
```
axelpezzo, line 2: Is this an import mistake or are you gonna use in somewhere?

src/components/accessories/patientTherapy/PatientTherapy.tsx
```diff
@@ -4,123 +4,139 @@ import TherapyForm from "./therapyForm/TherapyForm";
...
+      {(createStatus === "FAIL" || updateStatus === "FAIL") && (
+        <div ref={infoBoxRef} className="info-box-container">
+          <InfoBox type="error" message={t("common.somethingwrong")} />
+        </div>
+      )}
+
       <ConfirmationDialog
-        isOpen={hasSucceeded}
-        title="Therapy Created"
+        isOpen={createStatus === "SUCCESS" || updateStatus === "SUCCESS"}
```
axelpezzo, line 123: I would handler this behaviour with a useState: use an useEffect in order to properly update the state once one of this variables will change, do you agree?

axelpezzo, line 123: I would handler this behaviour with an useState: use an useEffect in order to properly update the state once one of this variables will change, do you agree?

src/libraries/formatUtils/dataFormatting.ts
tsognong: To handle timestamp formatting we should use moment(+date).format("format string). So if we want to ignore date string we should use 
`
return moment(+date).isValid()
    ? moment(+date).format("DD/MM/YYYY")
    : "";
`

src/libraries/reduxUtils/convert.ts
tsognong: We have two cases: millisecondes and date string. invalid date is also possible.

src/components/accessories/patientTherapy/therapyForm/TherapyForm.tsx
tsognong: mistake

src/components/accessories/patientTherapy/PatientTherapy.tsx
tsognong: May be we can do that. But the useSelector is a hook. Is it not duplicating the logic?

src/libraries/formatUtils/dataFormatting.ts
tsognong: We have to map each **disease code** gives by the form to the correspondant object.

src/libraries/formatUtils/dataFormatting.ts
tsognong: May be we should use return moment(+date).isValid()
    ? moment(+date).format("DD/MM/YYYY")
    : ""; As we are working with timestamp

src/components/accessories/patientTherapy/therapyForm/TherapyForm.tsx
tsognong: OK

src/libraries/formatUtils/dataFormatting.ts
tsognong: The disease returned by formik is just a string code, we have to convert it to the entire object in order to call the api. That is why I am using the + operator to convert that string to integer in order to find the right object.

src/libraries/reduxUtils/convert.ts
tsognong: OK, I changed it

src/libraries/formatUtils/dataFormatting.ts
axelpezzo: Is this already done?

axelpezzo (approved):

src/libraries/formatUtils/dataFormatting.ts
tsognong: OK, I applied since.

PR #172, opened 2021-08-26
Fix UI/menu and tab

tsognong:


src/components/activities/patientDetailsActivity/PatientDetailsActivity.tsx
```diff
@@ -204,9 +248,40 @@ const PatientDetailsActivity: FunctionComponent<TProps> = ({
...
+                      <div
+                        className={
+                          "align__element patientDetails__main_menu__item " +
+                          isActive("exams")
+                        }
+                        onClick={() => {
+                          setUserSection("exams");
+                          setDefaultRoute("/summary");
+                        }}
+                        style={{ display: "flex", alignItems: "center" }}
```
axelpezzo, line 284: Why are you gonna use the display flex and the class align__element?

I remember that when I release this code the alignaments of both icons and text were right.

tsognong: It is to center the text and the icon if I remember. So if you have some tips to align text with icon, you are welcome!!!!

tsognong: It is to center the text and the icon if I remember. So if you have some tips to align text with icon, you are welcome!!!!

axelpezzo: image

What I'm saying is that in my local branch the icons are already aligned with the text without any other style or css class.

tsognong: OK, I removed the flex and aligment style. It could be browser incompatibilty.

PR #178, opened 2021-09-06
Feature/op 646 check admitted patient

tsognong:


src/libraries/reduxUtils/convert.ts
```diff
@@ -1,17 +1,44 @@
...
+        obj[field] = item[field];
+      }
+      return obj[field];
+    });
+    return obj;
+  };
+  return data.map((item: any) => {
+    return itemRender(item);
+  });
+};
```
axelpezzo, line 44: I'm not sure about this. This method is both not generic and not specific. Are you trying to make a generic function (using any) but if you go ahead with the workflow I see some specific property (.description). I think that we could add some type in order to make this method works as expected with typescript.

tsognong: You are right. I use a combination of type there. It's OK.

PR #182, opened 2021-09-13
Refactoring/op 653 admission

tsognong:


src/components/accessories/admission/PatientAdmission.tsx
```diff
@@ -86,18 +102,58 @@ const PatientAdmission: FC = () => {
...
+        new Date(+(currentAdmission?.admDate ?? "")),
+        new Date()
+      ).toString(),
+      type: "number",
+    },
+    transUnit: {
+      value: currentAdmission?.transUnit?.toString() ?? "",
+      type: "text",
+    },
+  };
```
axelpezzo, line 138: I think that we could move this inside a const.ts file

axelpezzo: or config

tsognong: May be a config. We cannot use a const since we have to use the currentAdmission which is dynamic data.

axelpezzo: Yeah you right, config sounds great

axelpezzo:
@tsognong sounds OK, please resolve the conflicts in order to fix the merge.

PR #183, opened 2021-09-15
Feature/op 657 search bills and table

tsognong:


src/components/accessories/autocompleteField/AutocompleteField.tsx
```diff
@@ -43,14 +46,27 @@ const AutocompleteField: FunctionComponent<IProps> = ({
...
+      (event: any, value: any) =>
+        onInputChange ? onInputChange(event, value) : null,
+      250
+    ),
+    []
+  );
+
+  const actualClassName =
+    theme === "light" ? "autocomplete__light" : "autocomplete";
+
```
axelpezzo, line 60: I don't understand this, what's the purpose of the theme props?

src/components/accessories/billFilterForm/BillFilterForm.tsx
```diff
@@ -0,0 +1,167 @@
...
+
+  const validationSchema = object({
+    toDate: string().test({
+      name: "toDate",
+      message: t("bill.validatetodate"),
+      test: function (value) {
+        return +value >= +this.parent.fromDate;
+      },
+    }),
+  });
```
axelpezzo, line 51: I think that here we can exapand the validationSchema requiring all the fields as mandatory and add some validation on the date format values.

src/components/accessories/billFilterForm/BillFilterForm.tsx
```diff
@@ -0,0 +1,167 @@
...
+                errorText={getErrorText("patient")}
+                onBlur={onBlurCallback("patient")}
+                isLoading={searchStatus === "LOADING"}
+                options={renderOptions(patientSearchResults)}
+                onInputChange={handlePatientSearchChange}
+                freeSolo={true}
+              />
+            </div>
+          </div>
+        </form>
```
axelpezzo, line 161: I think that we should add a button in order to submit the form

src/components/accessories/billFilterForm/BillFilterForm.tsx
```diff
@@ -0,0 +1,167 @@
...
+    setSearchPatientParams({
+      id: !isNaN(+value) ? value : "",
+      firstName: isNaN(+value) ? value : "",
+      secondName: "",
+      birthDate: "",
+      address: "",
+    });
+    dispatch(searchPatient(searchPatientParams));
+  };
+
```
axelpezzo, line 42: Actually, I don't understand what's the puropose of this. Why +value is taken as id and firstName?

src/components/accessories/billFilterForm/consts.ts
```diff
@@ -0,0 +1,33 @@
...
+  },
+  fromDate: {
+    value: new Date().setHours(0) + "",
+    type: "date",
+  },
+  toDate: {
+    value: new Date().setHours(23) + "",
+    type: "date",
+  },
+};
```
axelpezzo, line 33: Where are we gonna use all of this fields?

src/components/accessories/billFilterForm/consts.ts
```diff
@@ -0,0 +1,33 @@
...
+    value: "",
+    type: "text",
+    options: [],
+  },
+  fromDate: {
+    value: new Date().setHours(0) + "",
+    type: "date",
+  },
+  toDate: {
+    value: new Date().setHours(23) + "",
```
axelpezzo, line 30: There is a reason why you are using setHours?

src/components/accessories/billTable/BillTable.tsx
```diff
@@ -0,0 +1,137 @@
...
+        break;
+      case "CLOSE":
+        dispatch(searchBills(fromDate, toDate, patientCode));
+        break;
+      case "DELETE":
+        dispatch(searchBills(fromDate, toDate, patientCode));
+        break;
+      case "ALL":
+        dispatch(searchBills(fromDate, toDate, patientCode));
+        break;
```
axelpezzo, line 53: case "ALL" + default

Plus, is because API'ìs are not ready yet that you call the same method?

axelpezzo, line 53: I think that this could be all under the "default" case.

src/components/accessories/billTable/BillTable.tsx
```diff
@@ -0,0 +1,137 @@
...
+        };
+      });
+    if (status === "CLOSE") {
+      results.filter((item) => item.status === "C");
+    }
+    if (status === "DELETE") {
+      results.filter((item) => item.status === "D");
+    }
+    return results;
+  };
```
axelpezzo, line 88: I whish to put this under a useEffect statement in order to filter the results when the status variable changes.

So you won't filter / format data every time the table renderds but only when the status and the data will do.

Plus, the data format I will put inside a lib.ts in order to keep this useEffect as much clean as possibile.

src/components/accessories/billTable/RenderBillDetails.tsx
```diff
@@ -0,0 +1,235 @@
...
+    paddingRight: 4,
+    paddingLeft: 5,
+  },
+}));
+
+const currencyFormat = (num: number | undefined) => {
+  return num
+    ? "$" + num.toFixed(2).replace(/(\d)(?=(\d{3})+(?!\d))/g, "$1,")
+    : "";
+};
```
axelpezzo, line 44: Few things:

- Put this under /libraries/formatUtils.ts
- this function should accept number and string
- use Numbro (https://numbrojs.com/) in order to accept different currency format (now it is dollar but maybe could be something else in future)

src/components/accessories/billTable/styles.scss
```diff
@@ -0,0 +1,22 @@
...
+        margin-left: 0; 
+        padding-left: 0;
+    }
+    li span .MuiTypography-h5 {
+        padding-left: 0;
+        margin-left: 0;
+        display: inline;
+        text-decoration: none;    
+        color: #000; /* Fallback for older browsers */
+        color: rgba(0, 0, 0, 0.7);
```
axelpezzo, line 15: If in 2021 there is someone that still use IE 9 he / she didn't deserves this fallback :)

In every case, if you need something like that you could use some RGB color as for eg: #333232

src/components/accessories/billTable/types.ts
```diff
@@ -0,0 +1 @@
+export const totalRow = 10;
```
axelpezzo, line 1: if this is the only thing that this file contains, it should be renamed const.ts

src/components/accessories/customModal/CustomModal.tsx
```diff
@@ -0,0 +1,79 @@
...
+  const top = 50;
+  const left = 50;
+  return {
+    top: `${top}%`,
+    left: `${left}%`,
+    transform: `translate(-${top}%, -${left}%)`,
+    display: "inline-block",
+    width: "auto",
+  };
+}
```
axelpezzo, line 18: If top and left are static you could add this in a scss file.

src/components/accessories/customModal/CustomModal.tsx
```diff
@@ -0,0 +1,79 @@
...
+  },
+}));
+
+interface ICustomModal {
+  title: string;
+  description: string;
+  content: React.ReactElement;
+  open: boolean;
+  onClose: () => void;
+}
```
axelpezzo, line 39: Put this in a types.ts file.

src/components/accessories/customModal/styles.scss
```diff
@@ -0,0 +1,13 @@
+.makeStyles-paper-6 {
```
axelpezzo, line 1: makeStyles-paper-6 is not a unique class, I mean this could change in future and lead to a unhandled behaviours.

Please, use a specific class name as for eg "custom_modal" or something like this.

src/components/activities/manageBillActivity/ManageBillActivity.tsx
```diff
@@ -0,0 +1,288 @@
...
+  };
+  const dispatch = useDispatch();
+
+  const [expanded, setExpanded] = useState<string | false>("form");
+  const [isOpen, setIsOpen] = useState(false);
+  const [fromDate, setFromDate] = useState(
+    new Date().setDate(new Date().getDate() - 6) + ""
+  );
+  const [toDate, setToDate] = useState(new Date().setHours(23) + "");
+  const [patientCode, setPatientCode] = useState(0);
```
axelpezzo, line 45: I think that you could hanlde this as an object inside a single useState statement, I think this could be more readable.

new Date().setHours(23) -> why setHours?

src/components/accessories/billFilterForm/BillFilterForm.tsx
tsognong: Patient field is  not mandatory to filter bills,  filter dates have default values. And those defaults values help to avoid empty data when the user reaches the page.

src/components/accessories/billFilterForm/consts.ts
tsognong: Ok. I removed unused fields.

src/components/accessories/billFilterForm/BillFilterForm.tsx
axelpezzo: Ok I see for the patient.

But what happens if I delete the Date from field's value?

image

I think that both dates should be mandatory and validate as valid date

src/components/activities/manageBillActivity/ManageBillActivity.tsx
```diff
@@ -0,0 +1,288 @@
...
+  ];
+
+  const submit = (filter: any) => {
+    setFromDate(filter.fromDate);
+    setToDate(filter.toDate);
+    setPatientCode(filter.patient);
+  };
+
+  const handleOnExpanded = (section: string) => {
+    setExpanded(section === expanded ? "form" : section);
```
axelpezzo, line 142: setExpanded(section === expanded ? **"form"** : section);

should return false instead

src/components/activities/manageBillActivity/ManageBillActivity.tsx
```diff
@@ -0,0 +1,288 @@
...
+                        </AccordionSummary>
+                        <AccordionDetails>
+                          <BillFilterForm
+                            className="searchBills__formData__item"
+                            onSubmit={submit}
+                            fields={initializeBillFilter(fromDate, toDate)}
+                          />
+                        </AccordionDetails>
+                      </Accordion>
+                      <Accordion expanded={true}>
```
axelpezzo, line 214: Should be: 

`expanded={expanded === "details"}>`

src/components/accessories/billFilterForm/BillFilterForm.tsx
tsognong: It think it is right. I add it

src/components/accessories/billTable/BillTable.tsx
axelpezzo: There is something wrong with the searchBills action: when I acess to ALL tab the table don't display the data. It displays only when I switch betweens the tabs. Maybe my previous suggestion could lead to solve this, in any case just keep the eyes on this issue.

src/components/accessories/billTable/BillTable.tsx
```diff
@@ -0,0 +1,137 @@
...
+          patient: (
+            <Link
+              to={`/details/${item.billDTO?.patientDTO?.code}/edit`}
+              style={{ textDecoration: "none" }}
+            >
+              <strong>{item.billDTO?.patName}</strong>
+            </Link>
+          ),
+          amount: item.billDTO?.amount,
+          balance: item.billDTO?.balance,
```
axelpezzo, line 77: What is the difference between those two (I'm just asking)?

In any case if they are currency they should be formatted as before.

src/components/activities/manageBillActivity/styles.scss
```diff
@@ -0,0 +1,259 @@
...
+  }
+
+  .footer {
+    position: relative;
+  }
+    
+  .active {
+    background-color: lighten($c-gray, 10);
+  }
+
```
axelpezzo, line 258: There is some padding that we miss inside the accordion:

image

src/components/activities/newBillActivity/NewBillActivity.tsx
```diff
@@ -0,0 +1,34 @@
...
+  const userCredentials = useSelector<IState, TUserCredentials>(
+    (state) => state.main.authentication.data
+  );
+  return (
+    <div className="new_Bill">
+      <AppHeader
+        userCredentials={userCredentials}
+        breadcrumbMap={breadcrumbMap}
+      />
+      <SkeletonLoader />
```
axelpezzo, line 28: Put the Skeleton inside a Panel component in orde to the CSS Grid works.

src/components/accessories/billFilterForm/BillFilterForm.tsx
tsognong: I removed it. The idea here was to let user find user by typing the patient code also.

src/components/accessories/billFilterForm/consts.ts
tsognong: Yes. For filter purpose I set the time to 23 in order extend the time period.

src/components/accessories/billTable/BillTable.tsx
tsognong: You are right for that.

src/components/accessories/billTable/BillTable.tsx
tsognong: I resolved the problem. I created another reducer for pending bills.

src/components/accessories/billTable/RenderBillDetails.tsx
tsognong: OK

src/components/accessories/billTable/styles.scss
tsognong: OK, I applied it

src/components/accessories/billTable/types.ts
tsognong: OK, I removed it.

src/components/accessories/customModal/CustomModal.tsx
tsognong: I applied it

src/components/accessories/customModal/CustomModal.tsx
tsognong: I applied it

src/components/accessories/customModal/styles.scss
tsognong: You are right I removed it

src/components/activities/manageBillActivity/ManageBillActivity.tsx
tsognong: Right. But I removed  the expanded, because user should see both sections without any action

src/components/accessories/billTable/BillTable.tsx
tsognong: OK, I applied the currency format

src/components/activities/manageBillActivity/ManageBillActivity.tsx
tsognong: Ok.

src/components/activities/manageBillActivity/ManageBillActivity.tsx
tsognong: Right. But I removed the expand feature for the moment, since the user can directly see both sections.

src/components/activities/manageBillActivity/styles.scss
tsognong: OK

src/components/accessories/billFilterForm/BillFilterForm.tsx
axelpezzo: Ok maybe we can go though about. Maybe we can reuse this logic and instead of using a textfield we can filter the users by typing name or surname or code and use an autocomplete field in order to show them.

This sounds like a new Jira task for me, maby just poolish your code and then I open a new task.

src/components/accessories/billFilterForm/consts.ts
axelpezzo: Ok, but why? I think that the date at 24pm should be enought, right? I don't understand why you decrease the hours to 23

src/components/activities/manageBillActivity/ManageBillActivity.tsx
axelpezzo: Ok, your point it's right. Maybe we can expand the Accordion logic in order to set a bunch of accordions open? 

Plus, in a future, I will wish to have a section called Billing data that takes all the billing information and it shows them inside some report as graphics, tables, ecc... but for now what you did it's OK.

axelpezzo:
@almeida1492 this merge request is reviewed and ready to be merged, do you wanna take a look?

PR #196, opened 2021-10-15
Removed view details from patient exam table

tsognong:


vim.exe.stackdump
```diff
@@ -0,0 +1,5 @@
+Stack trace:
+Frame     Function  Args
+2002FE38  6401E3DA (0000034C, 0000EA60, 000000A4, 2002FEA8)
+2002FF68  640E5166 (2002FF96, 00000099, 00000000, 00000000)
+200300CC  640E1C2B (00000000, 00000000, 00000000, 00000000)
```
almeida1492, line 5: I think we can remove this file.

almeida1492 (changes requested):

PR #204, opened 2021-11-08
Feature/op 685 new bill form patient activity

tsognong:


src/components/accessories/patientNewBill/styles.scss
```diff
@@ -0,0 +1,118 @@
...
+      }
+      @include susy-media($smartphone) {
+        width: 1fr;
+      }
+    }
+    & > .patientNewBill_right {
+      display: flex;
+      flex-flow: column;
+      align-items: center;
+      background-color: hsl(2, 98%, 53%, .3);
```
axelpezzo, line 46: This light red overlay it's not make me crazy. My suggestion here is to change the background with using a light gray -- changing the background of the two grey buttons too (maybe you could use the same color for all the button),

src/components/accessories/patientNewBill/NewBillSide.tsx
```diff
@@ -0,0 +1,49 @@
...
+          {t("bill.pay")}
+        </SmallButton>
+      </div>
+      <div className="lg-divider"></div>
+      <div>
+        <TextButton onClick={() => {}}>{t("button.suspend")}</TextButton>
+      </div>
+      <div>
+        <TextButton onClick={() => {}}>{t("button.delete")}</TextButton>
+      </div>
```
axelpezzo, line 44: Just for now, put a trash icon inside the delete button and a stop / wait icon inside the suspend button.

src/components/accessories/patientNewBill/styles.scss
axelpezzo: Plus, there is something wrong about and the responsiveness (I think there is something wrong inside the flex options).

src/components/accessories/patientNewBill/NewBillSide.tsx
```diff
@@ -0,0 +1,49 @@
...
+      <div>
+        <SmallButton
+          onClick={() => {
+            handlePaymentDialog(true);
+          }}
+        >
+          {t("bill.pay")}
+        </SmallButton>
+      </div>
+      <div className="lg-divider"></div>
```
axelpezzo, line 38: Little UX trick: once the payment is complete, make the payment button disabled with a green check icon near the text "Payed".

src/components/accessories/patientNewBill/NewBillSide.tsx
axelpezzo: And I think that's fine to add the Refund button in order to undo a payment. 
Will be my task discuss with Ilario about his implementation and API.

src/components/accessories/patientNewBill/PatientNewBill.tsx
```diff
@@ -0,0 +1,81 @@
...
+            />
+            <div className="patientNewBill_buttons">
+              <SmallButton
+                onClick={() => {
+                  setShowItemPicker(true);
+                }}
+              >
+                {t("button.add")}
+              </SmallButton>
+              <SmallButton onClick={() => {}}>{t("button.save")}</SmallButton>{" "}
```
axelpezzo, line 49: Put the Add button to the bottom right of the table (it's just a UI trick, the call to action must stay to the right position).

And then, the put the save button inside the column with the rest of the buttons with the label "Save as draft"

src/components/accessories/patientNewBill/styles.scss
```diff
@@ -0,0 +1,118 @@
...
+@import "../../../../node_modules/susy/sass/susy";
+
+.patientNewBill {
+  display: flex;
+  flex-flow: column;
+  border: 1px solid hsl(2, 98%, 53%, .075);
+  border-radius: 8px;
+  box-shadow: 2px 2px 2px hsl(2, 98%, 53%, .075);
+  padding: 16px 8px 16px 8px;
+  row-gap: 1em;
```
axelpezzo, line 11: We already have the "overlay" effect inside the table, we don't need to use another wrapper with the same style.

src/components/accessories/patientNewBill/styles.scss
```diff
@@ -0,0 +1,118 @@
...
+    & > .patientNewBill_right {
+      display: flex;
+      flex-flow: column;
+      align-items: center;
+      background-color: hsl(2, 98%, 53%, .3);
+      border-radius: 8px;
+      padding: 1.5em .25em 1.5em .25em;
+      row-gap: 1em;
+      .smallButton, .textButton {
+        width: 10em;
```
axelpezzo, line 51: It's not a best practice use "em" as unit of measure of height, width or position. It's better use px and % for this.

src/components/accessories/patientNewBill/styles.scss
```diff
@@ -0,0 +1,118 @@
...
+      .xs-divider, .lg-divider {
+        height: 1px;
+        background-color: black;
+      }
+      .xs-divider {
+        width: 3.75em;
+      }
+      .lg-divider {
+        width: 11.5em;
+      }
```
axelpezzo, line 77: It's not good for the responsiveness the usage of static width/height like this. There is no need to do that if you did use the flex rightly inside the parent element.

src/components/accessories/patientNewBill/itemsTable/BillItemsTable.tsx
```diff
@@ -0,0 +1,44 @@
...
+interface IOwnProps {
+  handleEdit: (row: any) => void;
+  handleDelete: (row: any) => void;
+}
+
+const BillItemsTable: FC<IOwnProps> = ({ handleEdit, handleDelete }) => {
+  const { t } = useTranslation();
+  const label = {
+    type: t("bill.type"),
+    description: t("bill.description"),
```
axelpezzo, line 16: I think that we should rename the label in "Type" and "Description" (now it's Bill.Type and Bill.Description).

src/components/accessories/patientNewBill/itemsTable/BillItemsTable.tsx
```diff
@@ -0,0 +1,44 @@
...
+    qty: t("bill.quantity"),
+    cost: t("bill.amount"),
```
axelpezzo, line 18: I think that we should use "Total amount" or "Amount per item" based if the description is referred to the total amount (quantity * single_price) or the price of a single item.

src/components/accessories/patientNewBill/styles.scss
SteveGT96: So we have to remove the borther and the elevation effect; right ?

axelpezzo: Yep

axelpezzo:
Hello @tsognong is this branch ready to be merged? did you apply my suggestions?

src/components/accessories/patientNewBill/itemsTable/BillItemsTable.tsx
tsognong: We prefer to use just amount here in order to save space. There is no much space to display a long column label.

src/components/accessories/patientNewBill/itemsTable/BillItemsTable.tsx
tsognong: OK. we translated them.

src/components/accessories/patientNewBill/NewBillSide.tsx
tsognong: For the refund button, we keep it for the moment as you will have to discuss with Ilario

src/components/accessories/patientNewBill/styles.scss
tsognong: OK

tsognong:
> 

@axelpezzo  your suggestions are applied. I would like to mention that this branch is just to display the new billing form ui, not the logic.

PR #210, opened 2021-11-15
Op-695-payment-amount-with-decimal

tsognong:


src/components/accessories/paymentDialog/PaymentDialog.tsx
```diff
@@ -90,6 +81,14 @@ export const PaymentDialog = ({
       : "";
   };
 
+  const withValueLimit = (values: NumberFormatValues) => {
+    if (values.floatValue)
+      return values.floatValue <= initialValues.paymentAmount;
```
axelpezzo, line 86: I think that we should use the updated value of paymentAmount and not the Formik initialValues

tsognong: Hello @axelpezzo the initialValues.paymentAmount contains the total amount to pay and it should not change.

axelpezzo: Ok so if I made a partial payment (e.g. 500 on a total of 1200), when I reopen this modal I should see 700 right as paymentAmount right? If is it so ignore this comment.

PR #211, opened 2021-11-24
OP-708-global-billing-activity-refinement

tsognong:


src/components/accessories/billTable/styles.scss
```diff
@@ -39,13 +39,13 @@
...
-      width: 33.33%;
+      padding: 0px 3px;
+      width: 14.28%;
       @include susy-media($narrow) {
-        padding: 0px 10px;
+        width: 50%;
       }
       @include susy-media($tablet_land) {
-        padding: 0px 10px;
+        width: 50%;
```
axelpezzo, line 48: We're not gonna use static width/height for display inline element but we use the flex property.

axelpezzo: Plus, in small desktop, the filter's items content are hidden by an overflow. I suggest to rethink the alignament of the fields (maybe you could add some flex-column and then scale down at each breakpoint, this is just a suggestion).

image

src/components/accessories/billTable/BillTable.tsx
```diff
@@ -245,26 +235,51 @@ export const BillTable: FC<IBillTableProps> = ({
...
+                    theme="regular"
+                    format="dd/MM/yyyy"
+                    isValid={isValid("year")}
+                    errorText={getErrorText("year")}
+                    label={t("bill.year")}
+                    onChange={dateFieldHandleOnChange("year")}
+                  />
+                </div>
+                <div className="filterBillForm__item">
+                  <PatientPicker
```
axelpezzo, line 267: It seems like that the components doesn't work as expected, when you try to digit something nothing appears.

src/components/accessories/billTable/BillTable.tsx
```diff
@@ -20,37 +19,40 @@ import {
...
+  const header = [
+    "id",
+    "patient",
+    "patId",
+    "date",
+    "status",
+    "lastPayment",
+    "amount",
+    "balance",
+  ];
```
axelpezzo, line 44: This is a lot to show inside a responsive table. I think that you could move some coulmns inside the row,

src/components/accessories/billTable/BillTable.tsx
tsognong: This is the patient picker that I design before. You should click on the loop to open search dialog.

tsognong: It is just a suggestion. I designed a patientAutocomplete and a patient Picker, but here I use the  PatientPicker.

src/components/accessories/billTable/BillTable.tsx
tsognong: Ok

src/components/accessories/billTable/BillTable.tsx
axelpezzo: I think that the PatientAutocomplete is enough in this case.

Plus, the PatientPicker has some UI / UX bugs that I would like to discuss during the next call.

src/components/accessories/billTable/styles.scss
axelpezzo: It's pretty close to be OK, there is still something to do about the alignment. When you reduce the browser's window the fields' alignament is pretty funny. 

Proposal solutions, group the fields by:
4 date fields
2 other fields
1 button

one type for each row.

src/components/accessories/billTable/BillTable.tsx
```diff
@@ -245,23 +257,46 @@ export const BillTable: FC<IBillTableProps> = ({
...
+                    views={["year"]}
+                    fieldValue={formik.values.year}
+                    disableFuture={true}
+                    theme="regular"
+                    format="yyyy"
+                    isValid={isValid("year")}
+                    errorText={getErrorText("year")}
+                    label={t("bill.year")}
+                    onChange={dateFieldHandleOnChange("year")}
+                  />
```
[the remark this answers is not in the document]
axelpezzo: image

What is the puropose of this fields when we already have the date from and the date to filter?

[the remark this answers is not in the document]
tsognong: I rely on the Marco design. But it is just a shortcut to set  fromdate and todate values. I aggree with you that these fields are not so useful. But as I am applying the marco design, it will be good to discuss with Ilario for that.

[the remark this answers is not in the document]
tsognong: > image
> 
> This happens when I try to edit the date field manually

Ok I will fix it.

src/components/accessories/billTable/styles.scss
tsognong: Ok thanks for the suggestion

src/components/accessories/billTable/BillTable.tsx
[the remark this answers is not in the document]
axelpezzo: > I rely on the Marco design. But it is just a shortcut to set fromdate and todate values. I aggree with you that these fields are not so useful. But as I am applying the marco design, it will be good to discuss with Ilario for that.

For now just disabled those two fields and focus on the rest

PR #213, opened 2021-12-08
OP-718 billing dashboard improvement

tsognong:


src/components/accessories/billTable/BillTable.tsx
```diff
@@ -245,23 +248,46 @@ export const BillTable: FC<IBillTableProps> = ({
...
+                  <PatientPicker
                     theme={"regular"}
                     fieldName="patientCode"
                     fieldValue={formik.values.patientCode}
                     label={t("bill.patient")}
                     isValid={isValid("patientCode")}
                     errorText={getErrorText("patientCode")}
                     onBlur={onBlurCallback("patientCode")}
-                    freeSolo={true}
                   />
```
axelpezzo, line 288: I think that the UX of this component is not the best: I mean if I blur on the text field nothing happens. I discover that in order to search the patient I must click on the icon inside the text fields, this is not usable for the most of the user.

I suggest to open the modal once the user click (or blur) on the textField in order to be more usable and intuitive.

Then inside the modal, I think that the UI is great but I would change the rappresentation of the list's users: instead of have 3 cards per line I would show those cards once per line (like the patient search). In fact, I think you could reuse the same component.

axelpezzo: PLus the modal's header should be formatted like this: title align to left (or center) and close button to the right. (UX Trick: the cta action should always align to the right)

src/components/accessories/billsRecap/BillsRecap.tsx
```diff
@@ -0,0 +1,244 @@
...
+            false
+          )}
+        />
+        <Doughnut
+          data={getDataFromTwoObjects("currentUserCashIn", "currentUserDebt")}
+          options={getOptions(
+            t("bill.currentuser", { name: userCredentials?.displayName ?? "" }),
+            false
+          )}
+        />
```
axelpezzo, line 211: Just for now, try to remove this in order to have 4 and 4 chart.

src/components/accessories/billsRecap/BillsRecap.tsx
```diff
@@ -0,0 +1,244 @@
...
+        <Bar
+          options={getOptions(t("bill.mostindebtpatients"))}
+          data={getDataFromObject("mostIndebtedPatients", t("bill.totaldebt"))}
+        />
+        <Line
+          options={getOptions(t("bill.paymentsvariations"))}
+          data={paymentsVariationsData}
+        />
```
axelpezzo, line 240: Move this after the cashier chart.

src/components/accessories/billsRecap/BillsRecap.tsx
```diff
@@ -0,0 +1,244 @@
...
+          options={getOptions(t("bill.bestpatients"))}
+          data={getDataFromObject(
+            "bestPatientsByPayments",
+            t("bill.totalpayment")
+          )}
+        />
+        <Bar
+          options={getOptions(t("bill.mostindebtpatients"))}
+          data={getDataFromObject("mostIndebtedPatients", t("bill.totaldebt"))}
+        />
```
axelpezzo, line 236: Before this 4 charts, add a select list in order to select the year. Just for now put the select's values as static options like 2019/2020/2021/2022.

This value should be passed to the getDataFromObject in order to get only the data of that year.

Plus, for each label add the selected year.

src/components/accessories/billTable/BillTable.tsx
tsognong: OK

tsognong: > I think that the UX of this component is not the best: I mean if I blur on the text field nothing happens. I discover that in order to search the patient I must click on the icon inside the text fields, this is not usable for the most of the user.
> 
> I suggest to open the modal once the user click (or blur) on the textField in order to be more usable and intuitive.
> 
> Then inside the modal, I think that the UI is great but I would change the rappresentation of the list's users: instead of have 3 cards per line I would show those cards once per line (like the patient search). In fact, I think you could reuse the same component.

About the patient list in the modal, it is responsive, it goes from 3 to 2 and then 1 according to de device size.  why don't we do like that?

src/components/accessories/billsRecap/BillsRecap.tsx
tsognong: OK

src/components/accessories/billTable/BillTable.tsx
axelpezzo: The card si small, there are (or there could be) some text that outlines outside the card because is too long to handler. In a mobile first UI if we have some space to use, use it at all.

tsognong: Ok you are right

tsognong: I have done it.

src/components/accessories/billsRecap/BillsRecap.tsx
tsognong: I have done it

src/components/accessories/billTable/BillTable.tsx
tsognong: Ok, I applied it

src/components/accessories/billsRecap/BillsRecap.tsx
tsognong: Ok I applied it. It is a good suggestion. Also I added an empty infox box placeholder when the selected data does'nt have  bills.

tsognong:
This PR is ready

PR #227, opened 2022-01-24
start OP-756-patient picker look and feel and pagination

tsognong:


src/components/accessories/patientPicker/PatientPicker.tsx
```diff
@@ -61,7 +51,16 @@ const PatientPicker: FC<IProps> = ({
   const { t } = useTranslation();
   const dispatch = useDispatch();
   const inputRef = useRef<any>(null);
-
+  const [currentPage, setCurrentPage] = useState(1);
+  const [patientsPerPage] = useState(5);
```
axelpezzo, line 55: Consider to promote those values as constants in a dedicated file.

src/components/accessories/patientPicker/PatientPicker.tsx
```diff
@@ -252,6 +248,7 @@ const PatientPicker: FC<IProps> = ({
                     errorText={getErrorText("id")}
                     onBlur={formik.handleBlur}
                     type="number"
+                    disabled={searchStatus === "LOADING"}
```
axelpezzo, line 251: Move this control outside the JSX code, inside a variable in order to make this more maintenable.

And then, use the ES6 JS feature in order to pass the disable parameter without the esplicit assignament

axelpezzo: This is relative to each line `disabled={searchStatus === "LOADING"}`

src/components/accessories/patientPicker/PatientPicker.tsx
```diff
@@ -300,14 +300,39 @@ const PatientPicker: FC<IProps> = ({
...
+                  <CircularProgress
+                    style={{ marginLeft: "50%", position: "absolute" }}
+                  />
+                )}
               </div>
             </form>
-            <div className="patientSearchResult">{renderSearchResults()}</div>
+            <div className="patientSearchResult">
+              {renderSearchResults()}
+              {(patientData?.length ?? 0) > 0 && (
```
axelpezzo, line 324: I would add a searchStatus === SUCCESS, don't you?

src/components/accessories/patientPicker/PatientPicker.tsx
tsognong: Please can you give me some tips about the **ES6 JS feature ** your talking about?

src/components/accessories/patientPicker/PatientPicker.tsx
tsognong: Ok right

src/components/accessories/patientPicker/PatientPicker.tsx
tsognong: OK

src/components/accessories/patientPicker/PatientPicker.tsx
axelpezzo: If you have a disabled var you can just put "disabled" without assign a variable explicitly as prop.

tsognong: I tried it but it does'nt work. When we use disabled prop without assigment it will take the default value which is true.

tsognong: I was searching how to disable globally the form or use useFormikContext but it seems not possible

tsognong: So the solution now is just putting *disabled={isLoading}* on every fields

PR #228, opened 2022-01-26
Op 755 billing dashbaord loader

tsognong:


src/components/accessories/billsRecap/BillsRecap.tsx
```diff
@@ -323,70 +327,66 @@ export const BillsRecap: FC = () => {
...
+              options={getOptions(
+                t("bill.bestsellingbyquantity"),
+                true,
+                isMobile(width) ? 1 : 3,
+                year.value
+              )}
+              data={getDataFromObject(
+                "bestSellingByQuantity",
+                t("bill.totalquantity")
+              )}
```
axelpezzo, line 342: I would use a ternary condition here in order to show or a CircularProgress or the Bar component.

tsognong: May I missed your explanation on Jira.

tsognong: But when we replace the chart div totally by the loader it does'nt look nice. That's why I just prefer to show the loader over the bar div. What do you think?

axelpezzo: Yeah sounds fine and, actually, the result in the UI is good.

PR #231, opened 2022-01-28
Op 762 handle 401 api errors

tsognong:
At that ime I just redirect a user to login page when the token expires

### DIFF
--- a/src/libraries/authUtils/tokenHasExpired.ts
+++ b/src/libraries/authUtils/tokenHasExpired.ts
@@ -1,4 +1,5 @@
 import jwtDecode from "jwt-decode";
+import { TOKEN_EXPIRATION_TIMEOUT } from "../../consts";
 
 export interface JwtTokenModel {
   exp: number;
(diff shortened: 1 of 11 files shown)


axelpezzo:
@almeida1492 seems ok to me, what do you think?

src/libraries/authUtils/tokenHasExpired.ts
```diff
@@ -0,0 +1,13 @@
...
+
+export interface JwtTokenModel {
+  exp: number;
+  sub: string;
+  auth: string;
+}
+
+export const tokenHasExpired = (token: any): boolean => {
+  const { exp } = jwtDecode<JwtTokenModel>(token);
+  const expirationTime = exp * 1000 - 60000;
```
axelpezzo, line 11: I would move this in a constant

alefalezza: @SteveGT96 can you do that?

Move "60000" to a constant `TOKEN_EXPIRATION_TIMEOUT` in `src\consts.ts`

SteveGT96:
OK Alessandro !

src/libraries/authUtils/tokenHasExpired.ts
SteveGT96: @alefalezza It's done !

axelpezzo:
@almeida1492 can we go ahead with this MR?

almeida1492 (approved):

almeida1492:
> @almeida1492 can we go ahead with this MR?

I think so.

PR #232, opened 2022-01-31
OP-765 | Fix order and patient dashboard

axelpezzo:
@tsognong I've add a new prop to the Table component called dateFields in order to fix the order. In fact before, the table component always uses the dateComparator to understand the rows's order. Now, with this parameter, we should use the dateComparator only if the fields is inside the dateFields, viceversa the Table should use the defaultComparator.


tsognong:
I am currently testing this version in the interface. Tomorrow I will let you know.

src/components/accessories/patientSummary/patientSummaryByDate/PatientSummaryByDate.tsx
```diff
@@ -37,7 +37,7 @@ const PatientSummaryByDate: FunctionComponent<TProps> = ({
         {!isLoading ? (
           <Table
             rowData={renderSummary(summaryData, dateFields, labels, medicals)}
-            compareRows={dateComparator}
+            dateFields={dateFields}
```
tsognong, line 40: I check the code and also the ui and it OK for me. At least you should remove unused dateComparator import `import { dateComparator } from "../../../../libraries/sortUtils/sortUtils";`  from the files that you changed.

src/components/accessories/table/Table.tsx
```diff
@@ -23,19 +23,22 @@ import {
...
+} from "../../../libraries/sortUtils/sortUtils";
 import ConfirmationDialog from "../confirmationDialog/ConfirmationDialog";
 import { useTranslation } from "react-i18next";
 import warningIcon from "../../../assets/warning-icon.png";
 
 const Table: FunctionComponent<IProps> = ({
   rowData,
   tableHeader,
+  dateFields = [],
   labelData,
```
tsognong, line 38: But I was thinking if it is not possible to factorize all these property with just one array of objects like:
`fields= [ {name: xxxxx, header: true|false, date: true|false, order: 1/2/3... }, ...]` 

And now in the table component we just use the same fields array. We can use map function to extract all what we want. What you have done is ok too. What do you think about my suggestion?

tsognong (changes requested):

src/components/accessories/table/Table.tsx
axelpezzo: This is a great suggestion in order to make the code more simple and maintenaible. I'm gonna open a new Jira task.

tsognong: Ok you are right. It should be a new task.

PR #250, opened 2022-03-18
OP-825 | Move Exam form from Clinic to Laboratory

SteveGT96:


almeida1492 (changes requested):
In the **Patient Dashboard**, the **Laboratory** sheet should present just what's currently under the tab **Exams**. Thus, **Laboratory** shouldn't have a tab menu, given that there will not be another menu layer within it.

localhost_3000_details_1_exams

SteveGT96:
@almeida1492 I removed tabs according to your request.

almeida1492 (approved):

PR #253, opened 2022-03-24
OP-800 | Show only admission history for inpatient.

SteveGT96:


almeida1492 (changes requested):
The tabs should be removed because we **won't have a second layer of navigation within the Patient Dashboard**. The only content for Admissions when the patient is "inpatient" should be the sheet Admissions with the table of data. Nothing else.

localhost_3000_details_1_admissions

almeida1492 (approved):

PR #258, opened 2022-03-31
OP 797 | implement visit section

SteveGT96:


src/components/activities/patientDetailsActivity/PatientDetailsActivity.tsx
```diff
@@ -440,13 +434,22 @@ const PatientDetailsActivity: FunctionComponent<TProps> = ({
...
-                      <PatientDetailsContent
-                        title="Admissions"
-                        content={PatientAdmissions}
-                      />
-                    </div>
+                  ) : userSection === "visits" ? (
+                    patient?.data?.status === PatientDTOStatusEnum.O ? (
+                      <div className="patientDetails__nested_content">
+                        <PatientDetailsContent
+                          title="Visits"
```
axelpezzo, line 441: Title here should be OPD, right?

src/components/activities/patientDetailsActivity/PatientDetailsActivity.tsx
```diff
@@ -440,13 +434,22 @@ const PatientDetailsActivity: FunctionComponent<TProps> = ({
...
+                        />
+                      </div>
+                    ) : (
+                      <div className="patientDetails__nested_content">
+                        <PatientDetailsContent
+                          title="Visits"
+                          content={PatientVisit}
+                        />
+                      </div>
+                    )
```
axelpezzo, line 452: I can't understand why this components are printed inline and not inside the RouterTabs as same as the orher.

axelpezzo: Ok I see now, it's to avoid the shows of the tab when a tabs has only one item. But, I think you could move this logic inside the RouterTabs component instead of create those excepions. I think that if you are inside the RouterTabs component you could still avoid to show the tab if there is only one item.

SteveGT96: OK ! I'll do it !

almeida1492 (commented):
@SteveGT96 Did you use `build:api` to generate the API code?

src/components/activities/patientDetailsActivity/PatientDetailsActivity.tsx
almeida1492: Actually, the title is correct here.

SteveGT96: The title should be Visits both for inpatients and outpatients. But the content for outpatients will be the content of the OPD already implemented.

SteveGT96:
> @SteveGT96 Did you use `build:api` to generate the API code?

Yes I used that command !

almeida1492:
@SteveGT96 I see there's no API call once I filled the form and tried to submit the data. Is it missing because we don't have the endpoint or something?

src/components/activities/patientDetailsActivity/PatientDetailsActivity.tsx
axelpezzo: You're right! Maybe we could just remove the RouterTabs if we don't use it anymore and render the components inside the code.

almeida1492 (changes requested):
I see that the confirmation dialog is shown when the API returns a **200**, but it seems that if the call to `/oh-api/visit` fails (returning **400**), the UI doesn't show the error infoBox like it does in the other forms previously implemented. I think we could add this behaviour also to this one.

src/components/accessories/patientVisit/patientVisitForm/PatientVisitForm.tsx
```diff
@@ -50,6 +49,7 @@ const PatientVisitForm: FunctionComponent<TProps> = ({
     enableReinitialize: true,
     onSubmit: (values) => {
       const formattedValues = formatAllFieldValues(fields, values);
+      console.log(formattedValues);
```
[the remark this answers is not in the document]
SteveGT96: OK ! I'm still working on it. It's more easy for me to go ahead when you review that way. Thanks ! I'm gonna take a look for the 400.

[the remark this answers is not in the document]
almeida1492: But if you're still working on a PR, I'd ask you to mark it as a draft (like I showed you the other day, remember?).

almeida1492 (changes requested):
The app should display **only an error infoBox** if the POST request fails. That's the behavior applied to the forms already implemented (e.g. NewPatient, EditPatient). At the moment, it's displaying a dialog.

almeida1492 (approved):

src/components/activities/patientDetailsActivity/PatientDetailsActivity.tsx
```diff
@@ -441,26 +337,79 @@ const PatientDetailsActivity: FunctionComponent<TProps> = ({
...
+                          />
+                        )}
+                      </Route>
+                      <Route path={`${path}/visits`}>
+                        <PatientDetailsContent
+                          title="Visits"
+                          content={
+                            patient?.data?.status === PatientDTOStatusEnum.O
+                              ? PatientOPD
+                              : PatientVisit
```
axelpezzo, line 367: The right condition is inverse: if the patient is outpatient the app shows Visits instead shows OPD.

PR #259, opened 2022-03-31
OP-804 | Mark required fields

axelpezzo:


almeida1492 (changes requested):
We should not update the i18n resource files (but the default one) directly. Otherwise we're gonna have conflicts when Transifex tries to update them automatically. For editing those strings, we have to use the [Transifex app](https://www.transifex.com/informatici-senza-frontiere-onlus/openhospital-ui/dashboard/).

almeida1492:
> > We should not update the i18n resource files (but the default one) directly. Otherwise we're gonna have conflicts when Transifex tries to update them automatically. For editing those strings, we have to use the [Transifex app](https://www.transifex.com/informatici-senza-frontiere-onlus/openhospital-ui/dashboard/).
> 
> I updated the source strings on the Transifex App too, should it be enough right?

🤔  hmmm... maybe, yes.

almeida1492 (approved):

PR #262, opened 2022-04-02
OP-813 | Implement admission section

SteveGT96:


axelpezzo (approved):

almeida1492 (changes requested):
There's a discrepancy regarding the requirement here. If the patient status is "inpatient", the app should display the "Admissions" section containing the patient's admission history ([OP-800](https://openhospital.atlassian.net/browse/OP-800)). Otherwise, if the patient is "outpatient", then we should display the admittance form ([OP-813](https://openhospital.atlassian.net/browse/OP-813)).

Moreover, we need to solve those conflicts before merging.

almeida1492 (approved):

PR #265, opened 2022-04-04
OP-814 | Implement discharge section

SteveGT96:
Since diagnosis in fields have been limited to One in previous works, I've also limitted diagnosis out fields to One. I've seen that imageryCharge and cliDiaryCharge fields were mentionned in the discharge but not present in the DTO, so I added them in the admission DTO. Let me know if you agree with that, thanks.


src/components/accessories/discharge/dischargeForm/DischargeForm.tsx
```diff
@@ -0,0 +1,291 @@
...
+        name: "disDate",
+        message: t("admission.validatelastdate", {
+          admDate: moment(admission?.admDate ?? "").format("DD/MM/YYYY"),
+        }),
+        test: function (value) {
+          return moment(value).isSameOrAfter(moment(admission?.admDate ?? ""));
+        },
+      }),
+
+    disType: string().required(t("common.required")),
```
axelpezzo, line 85: Both fields are not showing the validation message if I submit the form without touch them.

axelpezzo: This is the same for the admission form , If i cleare the form and I try to submit it not all the messages show to the user.

SteveGT96: Let me check it out please !

almeida1492 (approved):

PR #273, opened 2022-04-08
OP-840 | Update error InfoBox

SteveGT96:


src/components/accessories/admission/PatientAdmission.tsx
```diff
@@ -47,6 +47,16 @@ const PatientAdmission: FC = () => {
     (state) => state.admissions.createAdmission.status
   );
 
+  const createError = useSelector<IState>(
```
almeida1492, line 50: Instead of casting the returning value, you can do something like the following:

```suggestion
  const createError = useSelector<IState, string>(
```

src/components/accessories/admission/PatientAdmission.tsx
```diff
@@ -104,7 +114,7 @@ const PatientAdmission: FC = () => {
       />
       {createStatus === "FAIL" && (
         <div ref={infoBoxRef} className="info-box-container">
-          <InfoBox type="error" message={t("common.somethingwrong")} />
+          <InfoBox type="error" message={createError} />
```
almeida1492, line 117: The identifier `createError` is not a function, so maybe we could avoid using a verb for naming that. I think `errorMessage` or something like that would be nice.

almeida1492 (changes requested):

PR #276, opened 2022-04-08
OP-843 | Set first item menu as Patient Dashboard default se…

SteveGT96:
…ction


almeida1492 (changes requested):
I see a misbehaviour here: If I open the Patient Dashboard, I see that the active menu item is **Admissions**, and that's correct. Then, if I click on another menu item and refresh the page, the menu will show **Admissions** as active, but the content will be the one from the previous section I opened. Also the route will not change properly back to `/details/1/admissions`.

SteveGT96:
> I see a misbehaviour here: If I open the Patient Dashboard, I see that the active menu item is **Admissions**, and that's correct. Then, if I click on another menu item and refresh the page, the menu will show **Admissions** as active, but the content will be the one from the previous section I opened. Also the route will not change properly back to `/details/1/admissions`.

OK @almeida1492 ! I'm gonna reproduce that and fix it. About your last words, would you like to be redirected to the first item menu(Admissions in this case) whenever you refresh the page no matter on what section of the patient dashboard you are ?

almeida1492:
> > I see a misbehaviour here: If I open the Patient Dashboard, I see that the active menu item is **Admissions**, and that's correct. Then, if I click on another menu item and refresh the page, the menu will show **Admissions** as active, but the content will be the one from the previous section I opened. Also the route will not change properly back to `/details/1/admissions`.
> 
> OK @almeida1492 ! I'm gonna reproduce that and fix it. About your last words, would you like to be redirected to the first item menu(Admissions in this case) whenever you refresh the page no matter on what section of the patient dashboard you are ?

@SteveGT96 I think the right behavior here would be: if the current route is `/details/1/visits` and the user refreshes the page, the app should re-open in the Visits section of the Patient Dashboard.

almeida1492 (approved):

PR #278, opened 2022-04-13
OP-847 | Fix datefield validation error display

SteveGT96:


src/components/accessories/dateField/DateField.tsx
```diff
@@ -45,10 +45,12 @@ const DateField: FunctionComponent<IProps> = ({
         format={format}
         id={fieldName}
         name={fieldName}
-        label={required === FIELD_VALIDATION.SUGGESTED ? label + " **" : label}
+        label={label}
```
almeida1492, line 48: @axelpezzo I'm wondering if the code removed here was your work on that "suggested-filling fields" feature. Is this change ok?

axelpezzo: Yep, we need to restore this. @SteveGT96

axelpezzo (approved):

PR #285, opened 2022-04-26
Update api staging3

axelpezzo:


SteveGT96 (changes requested):
This update affects particularly Bill and Laboratory. You can discard changes you made concerning bill and laboratory(actually you renamed field using the new api spec convention). We'll later fix that in task related to Bill or Exam since Billing and Exams are not  part of the current spring.
Actually e2e tests are failing just because exam e2e test has not been update with the change you made(date->examDate).
If @almeida1492  agrees, we could discard thoses changes and apply them in future tasks.
You can take a look here [#283](https://github.com/informatici/openhospital-ui/pull/283)

PR #287, opened 2022-04-29
OP-855 | Fix current known minor bugs

SteveGT96:


src/components/accessories/patientOperation/operationForm/OperationRowForm.tsx
```diff
@@ -74,7 +82,7 @@ const OperationRowForm: FC<OperationRowProps> = ({
   const dateFieldHandleOnChange = useCallback(
     (fieldName: string) => (value: any) => {
       setFieldValue(fieldName, value);
-      formik.validateField("opDate");
+      formik.setFieldTouched(fieldName);
     },
     [setFieldValue, initialValues.opDate]
   );
```
axelpezzo, line 88: I think that should be better to have a second callback which takes care of the onBlur method.

src/components/accessories/admission/consts.ts
```diff
@@ -12,7 +12,7 @@ export const initialFields: TFields<AdmissionFormFieldName> = {
     type: "number",
   },
   admDate: {
-    value: "",
+    value: new Date(Date.now()).toISOString(),
```
axelpezzo, line 15: the toISOString methond doesn't take care about the timezone, which could cause to some incongruences with the date. Please see what I do inside the libraries/formDataHandling/functions.ts:29 . Maybe we can extract the code at 32 - 35 and export it inside another method which we can use to get the right date inside the rest of the code.

PR #290, opened 2022-05-05
OH2-49 | Add Admissions section e2e tests

almeida1492:
This branch adds _e2e tests_ for the **Admissions** section of the **Patient Details Activity**. While doing this work, I noticed we call the endpoint `admissions/current` at the initialization of this section. It seems inappropriate though. If we are creating a new entry for **Admission**, we are not supposed to fill out the form fields with the data of a current **Admission** entry, because if a current **Admission** exists, we shouldn't see that form at all. I ask @SteveGT96 for feedback on this. If my guess is right, I'll open a task for fixing that surplus call.


SteveGT96 (changes requested):
I've removed /admissions/current call in the [#292 ](https://github.com/informatici/openhospital-ui/pull/292). So we can fix red pipline and merge.

PR #294, opened 2022-05-10
OP-788 | Reduce i18n options to 3

SteveGT96:


src/components/accessories/langSwitcher/LangSwitcher.tsx
```diff
@@ -2,38 +2,25 @@ import React, { FunctionComponent, useContext } from "react";
...
-  fr: "French",
-  es: "Spanish",
-  pt: "Portoguese",
-  ar: "Arabic",
-  sw: "Swahili",
-  am: "Amharic",
-  cs: "Czech",
-  sq: "Albanian",
-  zh: "Chinese",
-};
```
axelpezzo, line 19: I would leave this as fallback in case that the available languages inside the customisation file was empty.

Please handle this and then we can merge the code.

almeida1492 (approved):

PR #295, opened 2022-05-11
OP-869 | Implement admission edit

SteveGT96:


axelpezzo:
I think that there are more work to do here.

First, Admission is not only regards itselft but involves the Discarge data too. So, we should be able to edit both entities inside the same form.

Second, we need to strictly check the date. Actually, I can update an admission with a date newer that the discahrge 

image

axelpezzo (changes requested):
@SteveGT96 I think that there are more work to do here.

First, Admission is not only regards itselft but involves the Discharge data too. So, we should be able to edit both entities inside the same form.

Second, we need to strictly check the date. Now, I can update an admission with a date newer that the discahrge and this is wrong.

image

axelpezzo (changes requested):
We're close. I see that:

- The Ward fields doesn't work when I try to update and save the value (edit mode);
- The table should be ordered by "Discharge date" (the first item is the most recent discharge date);
- I added a subtitle in the orders form in order to be more explicative when you are in edit or insert mode, please add this item in the form too;
- I changed the message when the table is empty inside the other tables, please report that change in the admission table too.

PR #299, opened 2022-05-18
OH2-88 | Add 'Add Operation' in visit tabales, update operation row edit and create

SteveGT96:


axelpezzo (changes requested):
image

1. I think that the modal content could have a better alignament: the title, the subtitle ad the fields are not aligned well.

image

2. If I open the modal inside a smartphone I can't scroll down, all the window content is fixed as is and I can't for e.g. go down to press the save button. The expected behaviour is that I can scroll the modal but not the page behind.

SteveGT96:
> Why did you use a custom dialog component? I think that the existing modal component should be enough or no?

@axelpezzo the existing modal component doesn't have features I wanted to implement, and I did not want to change it since it's used by other components. That's why I've added another modal component that suits with my needs.

PR #300, opened 2022-05-23
OH2-16 | Use select field for the city field

SteveGT96:


src/components/accessories/patientDataForm/PatientDataForm.tsx
```diff
@@ -322,13 +324,15 @@ const PatientDataForm: FunctionComponent<TProps> = ({
           </div>
 
           <div className="patientDataForm__item">
-            <TextField
-              field={formik.getFieldProps("city")}
-              theme="regular"
+            <SelectField
```
axelpezzo, line 327: I think that Autocomplete should be better if I have more and more cities to find out.

PR #307, opened 2022-06-10
OH2-74 | Change login api

SteveGT96:


alefalezza:
Is the new api shared with backend team?

SteveGT96:
> Is the new api shared with backend team?

Yes ! I tested it with Arnaud work !

PR #309, opened 2022-06-21
OH2-98 | Implement filter form and data display, add missing state

SteveGT96:


almeida1492:
I've noticed an incongruence between the fields displayed in the table and the ones required in the task (sex, age, name status, etc are missing). We should adjust that before merging this PR.

almeida1492 (approved):

PR #319, opened 2022-07-18
OH2-80 | Change flow for new patient

SteveGT96:


almeida1492 (changes requested):
It looks like you forgot to apply the dialog component you wrote to the NewPatientActivity. When I submit the form, I still see the old confirmation dialog with just two buttons (keep editing and dashboard).

almeida1492 (changes requested):
If I click on **PATIENT DASHBOARD** after submitting the new patient form, the app is redirected to `/patients` while it should be redirected to `patient/details`.

Moreover, I see some funny styling in the `/patients` view. Do you see that too?

Screenshot 2022-07-26 at 14 31 07

SteveGT96:
> If I click on **PATIENT DASHBOARD** after submitting the new patient form, the app is redirected to `/patients` while it should be redirected to `patient/details`.
> 
> Moreover, I see some funny styling in the `/patients` view. Do you see that too?
> 
> Screenshot 2022-07-26 at 14 31 07

Hum sorry ! The devtools panel was opened and I did not notice that !

almeida1492 (changes requested):
The styling problem was solved, but I still see the wrong navigation at the end of the form submission.

> If I click on PATIENT DASHBOARD after submitting the new patient form, the app is redirected to `/patients`while it should be redirected to `patients/details/#id`.

almeida1492 (changes requested):
I was taking a look at the changes you introduced last and it looks like there's something still missing. If the user clicks on the button "PATIENT DASHBOARD", the app should redirect to `/patients/details/#id`, where `#id` is the id of the patient just created.

almeida1492 (approved):

PR #326, opened 2022-08-10
OH2-126 | Make ageType select reactive, bring exam laboratory form to the main content

SteveGT96:


almeida1492 (changes requested):
We should use the router to handle the laboratory activity navigation, just like in the patient activity. The expected behavior is: if I enter the `laboratory/new` URL, I'll see the new laboratory form view on the screen. Also, we have to update the breadcrumb nav. 

Screenshot 2022-08-22 at 10 05 19

almeida1492 (changes requested):
I've noticed that if the app is under the `/laboratory/new` route and we refresh the page, that's what comes out:

Screenshot 2022-08-23 at 09 59 59

almeida1492 (approved):

PR #329, opened 2022-08-11
Update docker environment

SteveGT96:


docker-compose.yaml
```diff
@@ -11,9 +11,6 @@ services:
       - 3030:3000
     volumes:
       - ./:/oh
-      - /oh/node_modules
```
alefalezza, line 14: Here I want node_modules to not be overwritten by mounting the local one, because we already installed in the build step.

alefalezza: Better solutions are appreciated!

docker/Dockerfile.frontend
```diff
@@ -1,12 +1,13 @@
...
 
 ENV PUBLIC_URL=/
 ENV REACT_APP_USE_MOCK_API=
 ENV REACT_APP_BASE_PATH=/
 
 VOLUME /oh
 WORKDIR /oh
-COPY ../package*.json /oh
+# COPY ../package*.json /oh/
+COPY . ./
```
alefalezza, line 10: This is not needed

alefalezza (approved):

alefalezza (approved):

PR #337, opened 2022-09-12
OH2-135 | Update login error message

SteveGT96:


src/components/activities/loginActivity/LoginActivity.tsx
```diff
@@ -111,7 +112,9 @@ const LoginActivity: FunctionComponent<TProps> = ({
                 hidden: status !== "FAIL",
               })}
             >
-              {t("login.incorrectcredentials")}
+              {errorMessage?.includes("too short")
+                ? t("login.passwordtooshort")
+                : t("login.incorrectcredentials")}
```
almeida1492, line 117: We have to alternate the error message according to some code inside the error object that defines the type of error we're dealing with. Using the error message string itself like that is not a good way to go because the message can be changed in the future and, in this case, our code will not work properly anymore.

SteveGT96: The request will return 400 error only for the short password ?

almeida1492: No. It's gonna be used for every bad request. We have to play with the error object that the API is gonna return within the response body. When the call goes 400, inside the response body we can have an object like `{ description: "PASSWORD_TOO_SHORT" }`. This constant can be defined by an enum in the yaml. So you don't have to write it manually inside the code, but use something like `errorDescriptionEnum.PASSWORD_TOO_SHORT`. If you declare it in the yaml and regenerate the code, you're gonna have access to that enum in the generated code

PR #340, opened 2022-10-04
OH2-142 | Add allergies and anamnesis opd form

SteveGT96:


almeida1492 (changes requested):
I think that if the fields "anamnesis" and "allergies" are empty, we could display a default message informing the user that no data was entered for those fields. Something like "_There are no registered allergies (or anamnesis) for this patient_".

almeida1492 (approved):

PR #357, opened 2022-11-04
OHCS-59 | Remove patient search cache during navigation

SteveGT96:


src/state/patients/reducer.ts
```diff
@@ -75,6 +76,13 @@ export default produce((draft: IPatientsState, action: IAction<any, any>) => {
       break;
     }
 
+    case SEARCH_PATIENT_RESET: {
+      draft.searchResults.status = "IDLE";
+      delete draft.searchResults.error;
+      delete draft.searchResults.data;
```
alefalezza, line 82: It would be safer to restore initial state, which is 
`draft.searchResults.data = []`

almeida1492 (approved):

PR #417, opened 2022-11-17
OHCS-62 | Navigate to home screen after login

SteveGT96:
Navigate to home screen after login when the user manually logged out previously


almeida1492 (changes requested):
I think we are not resetting the state after the log-out yet. I mean, if I log in, search for a patient, log out and log in again, I see there are some remaining state values from the previous session. Perhaps that's what @mwithi was pointing out as an issue.

mwithi:
> I think we are not resetting the state after the log-out yet. I mean, if I log in, search for a patient, log out and log in again, I see there are some remaining state values from the previous session. Perhaps that's what @mwithi was pointing out as an issue.

I didn't test yet the PR, but I guess the whole state should be reset. My initial test was just an example.

almeida1492 (changes requested):
@SteveGT96 The state is not being reset yet. I see you declared an action creator named `resetAppState`, but you don't call it anywhere.

SteveGT96:
> @SteveGT96 The state is not being reset yet. I see you declared an action creator named `resetAppState`, but you don't call it anywhere.

I forgot to apply it ! But I just did it right now.

almeida1492 (approved):

PR #434, opened 2023-01-12
OHCS-71 - OHCS-72 - OHCS-75 | Therapy cumulative defect fix

SteveGT96:


src/libraries/formatUtils/formatDateDiff.ts
```diff
@@ -0,0 +1,28 @@
...
+
+/**
+ * @param date1 The lower date
+ * @param date2 The upper date
+ * @param labels The array containing the period labels in the order "yearLabel, monthLabel, weekLabel, dayLabel"
+ * @returns Returns the diff between [date1] and [date2] formatted in "nbYears labels[0], nbMonths labels[1], nbWeeks labels[2], nbDays labels[3]"
+ */
+export const formatDateDiff = (date1: any, date2: any, labels: any[]) => {
+  const b = moment(date1);
+  const a = moment(date2);
```
axelpezzo, line 11: I would check if those dates are valid dates before to continue

src/libraries/formatUtils/formatDateDiff.ts
```diff
@@ -0,0 +1,28 @@
+import moment from "moment";
+
+/**
+ * @param date1 The lower date
+ * @param date2 The upper date
+ * @param labels The array containing the period labels in the order "yearLabel, monthLabel, weekLabel, dayLabel"
+ * @returns Returns the diff between [date1] and [date2] formatted in "nbYears labels[0], nbMonths labels[1], nbWeeks labels[2], nbDays labels[3]"
+ */
+export const formatDateDiff = (date1: any, date2: any, labels: any[]) => {
```
axelpezzo, line 9: We should use a type check in order to allow only string as parameter.

axelpezzo (changes requested):
@SteveGT96 I added some comments to your code

axelpezzo (approved):

PR #437, opened 2023-01-18
OHCS-60 | fix triage save on enter

tsognong:


axelpezzo:
Hello @tsognong I have a question about this task: what is the issue that you are trying to fix? I mean the form should still have the keys navigation active on it in order to improve the usability and the accessibility of the application, why are you trying to alterate this behaviour?

tsognong:
Hello @axelpezzo, the form is submitted whenever the user presses the enter key. We want the form to submit only when the user clicks on the submit button. It is what I want to solve.

axelpezzo:
I think that the problem here is the input focus that should be applied to the first input of the form and not on th submit button.

I mean the wrong behaviour here is not the "enter" event itself but understand which component call this event and then fix it.

tsognong:
@axelpezzo  Actually we just have two possibilities: either we keep the actual behaviour and permit to submit the form whenever the user presses the enter key **or** we prevent the auto submit behaviour and let the user decides when to submit the form by clicking on the submit button. 
I suggest to keep the default behaviour as it is and  delete that pull request.

PR #438, opened 2023-01-19
OHCS-51 OHCS-53 | no data found + infobox

tsognong:


src/components/accessories/infoBox/InfoBox.tsx
```diff
@@ -2,11 +2,23 @@ import classNames from "classnames";
...
+      <h5>
+        {type === "info" ? (
+          <InfoIcon fontSize="small" />
+        ) : type === "warning" ? (
+          <WarningIcon fontSize="small" />
+        ) : (
+          <ErrorIcon fontSize="small" />
+        )}
+        &nbsp; {message}
+      </h5>
```
axelpezzo, line 21: I would prefer use a separate method in order to render the icon and a switch case in order to catch the conditions. Plus, if the type props has a finite status I would use a enum.

axelpezzo: The space before the message I would add it on the icon itselft with the css (e.g margin-right: 10px).

tsognong: Ok thanks for the suggestions. Could you check again please?

src/components/accessories/infoBox/types.ts
```diff
@@ -1,4 +1,10 @@
...
-  type: "warning" | "error";
+  type: "info" | "warning" | "error";
   message: string;
 }
+
+export enum InfoBoxType {
+  Info = "info",
+  Warning = "warning",
+  Error = "error",
+}
```
axelpezzo, line 10: I think that should be more clean if we remove the static string on the IProps interface and use instead the belowed enum.

tsognong: Ok. I have applied your changes. In order to avoid changing all the components to use the enum, I set the type to be enum or string, you can recheck.

PR #444, opened 2023-01-26
OHCS-85|lab section check result value - proc 2

tsognong:


axelpezzo (changes requested):
@tsognong please resolve conflicts

PR #447, opened 2023-01-30
OH2-173 | Dashboard minor fixes

SteveGT96:
Update charts layout and restore desktop date picker with de possibility to select a single date instead of range date.


mwithi (approved):
LGTM. Thanks a lot!

PR #448, opened 2023-01-31
Fix dashboard

axelpezzo:


mwithi:
Hi @axelpezzo, with @SteveGT96 we had also worked on two aspects:

- the actual dates used in the range should always be visible
- it must also be possible to select a single day with the calendar

Can you work on them?

src/components/accessories/dashboard/dashboardContent/filter/DashboardFilter.tsx
```diff
@@ -129,39 +129,35 @@ export const DashboardFilter: FC<IOwnProps> = ({ onPeriodChange }) => {
...
-            fieldValue={dateRange[0]?.toISOString() ?? ""}
-            format="dd/MM/YYY"
-            onChange={handleDateChange}
-            TextFieldComponent={(props) => (
-              <IconButton onClick={onIconClickHandler}>
-                <CalendarTodaySharp />
-              </IconButton>
-            )}
-            open={open}
-          />
```
SteveGT96, line 164: Hi @axelpezzo ! The DesktopRangePicker doesn't allow to pick the same date(I think) ! By removing the DateField, we won't be anymore able to select a single date(did you try it ?)

PR #449, opened 2023-02-03
Fix/patient exams infinite loading

tsognong:


src/components/accessories/patientExams/patientExamsTable/PatientExamsTable.tsx
```diff
@@ -114,7 +114,7 @@ const PatientExamsTable: FunctionComponent<IOwnProps> = ({
           <InfoBox type="warning" message={t("common.emptydata")} />
         </div>
       )}
-      {labStatus === "LOADING" && (
+      {labStatus === "IDLE" && (
```
axelpezzo, line 117: Uhm what's about the loading status? I think that the issue is inside the labStatus value and not how it's rendered.

src/state/laboratories/actions.ts
```diff
@@ -142,10 +139,6 @@ export const searchLabs =
         examName: query.examName,
         patientCode: !isNaN(query.patientCode) ? query.patientCode : undefined,
       })
-      .pipe(
-        switchMap((labs) => getRowsForPrint(labs)),
-        catchError((error) => of([]))
-      )
```
axelpezzo, line 148: Why do you remove this?

tsognong: The api schema has been updated in order to include Rows In response, so it is not necessary to get the rows again.

src/components/accessories/patientExams/patientExamsTable/PatientExamsTable.tsx
tsognong: I have committed without changing it, it is a mistake. I correct it now

PR #450, opened 2023-02-03
OH2-178 | Dashboard - age type label fix and filter fix

SteveGT96:
Show translation for age types in graphs, add possibility to select single date and update period computation.

### DIFF
--- a/api/oh.yaml
+++ b/api/oh.yaml
@@ -5113,18 +5113,18 @@ paths:
       operationId: getPatientCitiesUsingGET
       produces:
         - application/json
-      
+
       responses:
-        '200':
+        "200":
           description: OK
           schema:
             type: array
-            items: string 
-        '401':
+            items: string
+        "401":
           description: Unauthorized
-        '403':
+        "403":
           description: Forbidden
-        '404':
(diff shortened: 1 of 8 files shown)


src/resources/i18n/en.json
```diff
@@ -35,7 +35,7 @@
       },
       "weeks": {
         "current": "This week",
-        "previous": "Last Weeks",
+        "previous": "Last Week",
         "last2": "1-2 weeks",
         "last3": "1-3 weeks"
```
mwithi, line 40: These should also be translated

SteveGT96: @mwithi , Intesys uses transifex for translations; on the frontend side, all we need to to is to add english keys.

SteveGT96: I it's OK for you, let me know, so I mark this conversation as resolved.

mwithi: ops, sorry, I thought it was the code (I oversight it😆)
It's ok then! Thanks!

mwithi (approved):
LGTM. Thanks a lot!

PR #455, opened 2023-02-07
OH2 180 update age field computation

SilverD3:


src/components/accessories/patientDataForm/PatientDataForm.tsx
```diff
@@ -98,6 +99,50 @@ const PatientDataForm: FunctionComponent<TProps> = ({
     }))
   );
 
+  const allAgeTypes = useSelector(
+    (state: IState) => state.ageTypes.getAllAgeTypes.data
+  );
+
+  const getBirthDateAndAge = (
```
axelpezzo, line 106: Can we have an hook for this?

src/components/accessories/patientDataForm/PatientDataForm.tsx
```diff
@@ -98,6 +99,50 @@ const PatientDataForm: FunctionComponent<TProps> = ({
...
+    ageType: string,
```
axelpezzo, line 107: Can we have a enum or a deterministic string?

src/components/accessories/patientDataForm/PatientDataForm.tsx
```diff
@@ -98,6 +99,50 @@ const PatientDataForm: FunctionComponent<TProps> = ({
...
+    formattedValues: Record<string, any>
+  ): { birthDate: string; age: number } => {
+    if (ageType == "agetype") {
```
axelpezzo, line 110: I think that a switch case would be more appropriate for the code reading in this case

src/components/accessories/patientDataForm/PatientDataForm.tsx
SilverD3: I'm not getting you well. Do you suggest to create an hook like the existing one `useCityOptions` ?

SilverD3: Also, the function `getBirthDateAndAge()` is only used in `PatientDataForm` Component.

axelpezzo, line 106: Sorry, I try to explain my self properly: can we move this method under the libraries folder? Like under formDataHandling that seems to me the right place to store this function.

src/components/accessories/patientDataForm/PatientDataForm.tsx
SilverD3: Okay

src/libraries/formDataHandling/functions.ts
```diff
@@ -247,3 +249,64 @@ export const isFieldSuggested = (
 ) => {
   return formCustomization.suggestedFields.includes(fieldName);
 };
+
+export const getBirthDateAndAge = (
+  ageType: string,
```
axelpezzo, line 254: Can we provide an enum for this field?

axelpezzo (approved):

PR #456, opened 2023-02-07
OH2-179 | make get hospital public

tsognong:
See OH2-179.

Paired with https://github.com/informatici/openhospital-api/pull/214.

### DIFF
--- a/src/components/Private.js
+++ b/src/components/Private.js
@@ -8,9 +8,5 @@ export const Private = () => {
   useAuthentication();
   const { pathname } = useLocation();
 
-  return isAuthenticated() ? (
-    <Outlet />
-  ) : (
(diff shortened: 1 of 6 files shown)


axelpezzo (approved):

mwithi:
Hi, I suggest to always do merge with "squash" so on the develop branch we'll just see one merge commit and not all commits occurred on the source branch.

The downside is that the final commits history on develop is different from the branch one, so comparing again after the merge the same branch with the develop would result in the same number of commit and changes. To overcome this, one should always start a new branch from develop for new changes after the merge, rather than continuing adding more commits on the same branch. (In other words, when a branch is merged it is supposed to be "obsolete" and should be deleted)

PR #463, opened 2023-03-03
OHCS-97 visit function wrong text length

SilverD3:
- Put `disease` and `diseaseType` each on a new line
- Fix 'undefined initial value' issue on fields `disease` and `diseaseType`

### DIFF
--- a/src/components/accessories/opds/filter/styles.scss
+++ b/src/components/accessories/opds/filter/styles.scss
@@ -57,7 +57,7 @@
             @include susy-media($watch) {
               width: 100%;
             }
-            &.col-4 {
+            &.w-100 {
               width: 100%;
             }
           }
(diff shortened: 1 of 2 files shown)


SilverD3:
@almeida1492 Please review my PRs

src/components/accessories/opds/filter/styles.scss
```diff
@@ -57,6 +57,9 @@
             @include susy-media($watch) {
               width: 100%;
             }
+            &.col-4 {
+              width: 100%;
+            }
```
axelpezzo, line 62: This is not the proper way to use col-4 as semantic class.

I suggest to change the selector name in something more "exclusive" and then fix the width that shouldn't be 33%?

axelpezzo (approved):

mwithi (approved):
LGTM. Thanks a lot!

mwithi:
@axelpezzo I don't know why, having merged this PR broke the workflow, can you check please?
https://github.com/informatici/openhospital-ui/actions/runs/4949970784

PR #475, opened 2023-05-23
OHCS-98 | Limit supported languages regardless the translation files

SteveGT96:


axelpezzo:
Hello @SteveGT96 I saw the discussion on Jira. Instead of physically delete the files, cannot we just I mean if we already have a range of supported and ready-to-use languages, cannot we just disable the autodetection of those languages?

SteveGT96:
> Hello @SteveGT96 I saw the discussion on Jira. Instead of physically delete the files, cannot we just I mean if we already have a range of supported and ready-to-use languages, cannot we just disable the autodetection of those languages?

@axelpezzo , I've discussed about that with @mwithi ! Could you have a look on comments [here](https://openhospital.atlassian.net/browse/OHCS-98) ?

SteveGT96:
Hello @axelpezzo ! I think it's gtg. Please, could you take a look ?

axelpezzo (approved):

PR #478, opened 2023-06-02
OH2-146 | Implement per-role landing page

SteveGT96:
See OH2-146.

### DIFF
--- a/src/libraries/hooks/useLandingPageRoute.ts
+++ b/src/libraries/hooks/useLandingPageRoute.ts
@@ -1,37 +1,23 @@
 import { useMemo } from "react";
+import { landingPagePriority } from "../../customization/landingPage";
+import { defaultRoute } from "../../routes/const";
 import { TPermission } from "../../types";
 import { usePermissions } from "../permissionUtils/usePermissions";
 
 /* Custom hook that returns the landing page route based on the user's permissions.
- * @returns {string|null} The landing page route or null if no permission matched.
+ * @returns {string} The landing page route.
  */
 export const useLandingPageRoute = () => {
   const permissions = usePermissions();
 
-  // Define the possible permissions to match to routes.
-  const PERMISSIONS: Record<string, TPermission> = {
-    DASHBOARD: "dashboard.access",
-    PATIENT: "patient.access",
-    LABORATORY: "laboratory.access",
-    VISIT: "visit.access",
-  };
-
   // Calculate the landing page route based on the user's permissions.
-  const landingPageRoute = useMemo(() => {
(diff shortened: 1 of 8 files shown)


alefalezza:
In my opinion it's ok, but I would suggest to push configurability a little forward.
What if dashboard url and priority should be configurable in `/src/customization` folder?

Example: add a new configuration file in `/src/customization/landingPage`:
```typescript
type TLandingPagePriority = {
  permission: TPermission,
  route: string
}

// Landing page will be the first route matching permission,
// you can change priority rules by moving elements in this list
export const landingPagePriority = [
  {
    permission: "dashboard.access",
    route: "/dashboard"
  },
  {
    permission: "patient.access",
    route: "/patients"
  }
  // ... and so on
]
```

Then, in `src/libraries/hooks/useLandingPageRoute.ts`:

```typescript
// ...
// return the first route matching permissions
const landingPage = useMemo(
 () => landingPagePriority.find((element) => permissions.includes(element.permission)),
 [landingPagePriority, permissions]
);

return landingPage.route;
```

alefalezza (changes requested):
My suggestions in comments

src/libraries/hooks/useLandingPageRoute.ts
```diff
@@ -0,0 +1,38 @@
...
+    if (permissions.includes(PERMISSIONS.PATIENT)) {
+      return "/patients";
+    }
+    if (permissions.includes(PERMISSIONS.LABORATORY)) {
+      return "/laboratory";
+    }
+    if (permissions.includes(PERMISSIONS.VISIT)) {
+      return "/visits";
+    }
+    return null;
```
axelpezzo, line 33: It is not an error just a code reading improvement: switch from an If / Else case to a Switch case.

axelpezzo:
@SilverD3 upon the Alessandro comment, I will suggest to remove the successRoute property on the RedirectAfterLogin component because could be lead to some misunderstanding. I suggesto to create a defaultRoute in the outer const.ts file and use that constant as default on the useLandingPageRoute result.

SteveGT96:
Hello @alefalezza , @axelpezzo ! Sorry for replying late ! I've been on vacations. I did not know you didn't know. I just resolved your comments.

src/routes/const.ts
```diff
@@ -0,0 +1 @@
+export const defaultRoute = "/patients";
```
alefalezza, line 1: @SteveGT96 I would suggest to move this const to /customization/landingPage.ts, because it can be customized and is related to route configuration.

mwithi (approved):
Approving after peer reviews

PR #479, opened 2023-06-06
OP-1104 Fix session audit update upon logout from UI

mwithi:
See OP-1104.

- set `withCredentials: true`

This PR is paired with https://github.com/informatici/openhospital-api/pull/283

### DIFF
--- a/src/generated/runtime.ts
+++ b/src/generated/runtime.ts
@@ -104,6 +104,7 @@ export class BaseAPI {
             headers,
             body: body instanceof FormData ? body : JSON.stringify(body),
             responseType: responseType ?? 'json',
+            withCredentials: true,
         };
     }
 
@@ -159,6 +160,7 @@ export interface RequestOpts extends AjaxRequest {
     headers?: HttpHeaders;
     body?: HttpBody;
     responseType?: 'json' | 'blob' | 'arraybuffer' | 'text';
+    withCredentials?: boolean;
 }
 
 export interface ResponseOpts {


alefalezza:
@mwithi I suppose the problem is not related to axios configuration, but to server. Auth/session cookies should be secure and not accessible by javascript, by setting "Secure" and "HttpOnly" flags server side.
Then I would close this PR without merging it.

mwithi:
> @mwithi I suppose the problem is not related to axios configuration, but to server. Auth/session cookies should be secure and not accessible by javascript, by setting "Secure" and "HttpOnly" flags server side. Then I would close this PR without merging it.

@alefalezza I've set `HttpOnly` server side with the setting `server.servlet.session.cookie.http-only=true`, but still I'm finding quite everywhere that the requests from the frontend need to use `withCredentials` because we are using CORS

see:
https://github.com/informatici/openhospital-api/pull/283/files
https://stackoverflow.com/a/63500174
https://stackoverflow.com/a/61442764 (last paragraph)
https://github.com/marmelab/react-admin/issues/7019#issuecomment-999404082
https://stackoverflow.com/q/71688058

Therefore, `HttpOnly` is true that is needed to prevent XSS access, but in order to exchange cookies back and forth with the server, `withCredentials` seems still required to be set on **both server and client** because we are using cors and wihtout it cookies are not allowed to be sent.

see:
https://developer.mozilla.org/en-US/docs/Web/API/XMLHttpRequest/withCredentials

In otherwords, if I only set `HttpOnly` server side, the cookie is not stored in the browser nor sent back to server, despite correctly received.

**Please make sense to you?**

`server.servlet.session.cookie.secure=true` instead, would make the cookie to be sent only over HTTPS connection. 
Since we don't have HTTPS yet in the develop but we do in test environment, we can leave it disabled (commented)

see:
https://stackoverflow.com/a/52295544

axelpezzo:
@mwithi is this PR ready for you?

mwithi:
@reviewers we cannot wait anymore because we have the release this week, so I will merge this PR by the moment it seems to work properly and solves the linked issue.

PR #482, opened 2023-06-16
Fix issue on dashboard concerning random behavior when changing screen size (OH2-194)

SilverD3:
I fixed the issue on dashboard, that was causing a random behavior when user's screen change and sometimes when user navigate to the dashboard page


mwithi (approved):
Tested quickly and much better than before, thanks!

PR #483, opened 2023-06-19
OHCS-91 | Update login error message

SteveGT96:


mwithi:
Before it was showing always "Invalid username and password" regardless if the API server was up or down.

Now it does the same with message "Something went wrong, please again later" even if the API server is up but the problem is over the username (wrong, not existing).

Of course the error message should related to the problem.

image

SteveGT96:
> 

Hello @mwithi. This is shown when the error response dosn't return error message. We need a standard way for error messages. That is not done yet.

SteveGT96:
@mwithi I've update the error message accoding to this [note](https://openhospital.atlassian.net/browse/OHCS-91?focusedCommentId=15029) from Alex. Now if you see "Something whent wrong, it's prolly because the description of the error returned by the backend is undefined.

src/resources/i18n/en.json
```diff
@@ -1,4 +1,11 @@
 {
+  "errors": {
+    "somethingwrong": "Something went wrong, please retry later.",
+    "incorrectcredentials": "Invalid username or password",
+    "passwordtooshort": "Password too short",
+    "badrequest": "Bad reauest",
```
mwithi, line 6: suggestion
```
    "badrequest": "Bad request",
```

mwithi:
> @mwithi I've update the error message accoding to this [note](https://openhospital.atlassian.net/browse/OHCS-91?focusedCommentId=15029) from Alex. Now if you see "Something whent wrong, it's prolly because the description of the error returned by the backend is undefined.

Well, not exactly... at the moment the backend response for invalid username or password is:

{
    "timestamp": "2023-07-05T23:20:08.523+00:00",
    "status": 401,
    "error": "Unauthorized",
    "trace": "org.springframework.security.core.userdetails.UsernameNotFoundException: admin123 was not found...",
    "message": "Not Authorized",
    "path": "/auth/login"
}

so the response does not contain any "description" at all. Moreover, what @axelpezzo is naming "code" is indeed "status", and "description" is indeed "error", so I think you have to use the latter for the translations (while the message is what is expected to be, a _human readable message_ in English, non expected to be translated in the UI).

Instead, when the server is down, you have no response at all, so in that case the UI can show any message, and I think it should be something like "Service unavailable" or whatever for a no-response (i.e. not only the "description" is missing, but the whole response).

Anyway, why we don't just align UI with the actual API before to propose any change to the format?

SteveGT96:
> > @mwithi I've update the error message accoding to this [note](https://openhospital.atlassian.net/browse/OHCS-91?focusedCommentId=15029) from Alex. Now if you see "Something whent wrong, it's prolly because the description of the error returned by the backend is undefined.
> 
> Well, not exactly... at the moment the backend response for invalid username or password is:
> 
> { "timestamp": "2023-07-05T23:20:08.523+00:00", "status": 401, "error": "Unauthorized", "trace": "org.springframework.security.core.userdetails.UsernameNotFoundException: admin123 was not found...", "message": "Not Authorized", "path": "/auth/login" }
> 
> so the response does not contain any "description" at all. Moreover, what @axelpezzo is naming "code" is indeed "status", and "description" is indeed "error", so I think you have to use the latter for the translations (while the message is what is expected to be, a _human readable message_ in English, non expected to be translated in the UI).
> 
> Instead, when the server is down, you have no response at all, so in that case the UI can show any message, and I think it should be something like "Service unavailable" or whatever for a no-response (i.e. not only the "description" is missing, but the whole response).
> 
> Anyway, why we don't just align UI with the actual API before to propose any change to the format?

@mwithi I pushed another commit.

mwithi (approved):
LGTM. Thanks a lot!

PR #484, opened 2023-06-20
OH2-194 Changed the responsiveness behavior of the dashboard

SilverD3:
See OH2-194.

I changed the responsiveness behavior of the dashboard. Now, when a user add a widget, that widget is added for all breakpoints (screen sizes). Idem for when the user remove a widget.


mwithi (approved):
Tested and work as expected by the PR title

PR #486, opened 2023-06-26
OH2-204 | Align with paginated endpoints

SteveGT96:
See OH2-204.


src/components/accessories/laboratory/Exams.tsx
```diff
@@ -54,11 +54,20 @@ export const Exams: FC = () => {
     (state: IState) => state.laboratories
   );
 
+  useEffect(() => {
+    setFilter({ ...filter, page: page });
+  }, [page]);
```
almeida1492, line 59: Not including `filter` in the dependencies array could cause some consistency problems. A correct implementation of this effect function should be like the following:
```suggestion
  useEffect(() => {
    setFilter((prevState) => ({ ...prevState, page: page }));
  }, [page]);
```

almeida1492 (changes requested):

almeida1492 (approved):

PR #493, opened 2023-07-17
OH2-214 | Make all Clinic data readonly

SteveGT96:
See OH2-214.


gasp (approved):
I assume this is also enforced server-side

mwithi:
> I assume this is also enforced server-side

"Clinical" is a specific page, but the fields can be updated by the same user in another page

mwithi (approved):
LGTM

PR #498, opened 2023-07-19
OH2-220 | Add reset button in visit and laboratory filters

SteveGT96:
See OH2-220.

Also added the reset button in laboratory filter form.


SteveGT96:
Hello @mwithi ! I've also added the reset button in laboratory filter form.

SteveGT96:
Hello @mwithi ! I've also added the reset button in laboratory filter form.

mwithi (changes requested):
Hello @SteveGT96 it seems that the two filter pages have different l&f: Visits has the filter button on the left, Laboratory on the right. I would make the same, and the RESET button always on the same side (on the left if FILTER is at right-end or viceversa).

Also looks like the two actions behave in different way (see OH2-220)

mwithi:
Up!

mwithi (approved):
LGTM. Thanks a lot!

PR #499, opened 2023-07-20
OH2-221 | Update visits filter styles

SteveGT96:


mwithi (changes requested):
It doesn't seem the DiseaseType and Disease fields got more space (as requested in the issue), instead I see unrequested spaces between other patients' fields

image

SteveGT96:
Hello @mwithi ! I've tested with the Arnaud's backend. Dates are showing correctly. Arnaud just need to perform check for the filter that's returning no content.

mwithi:
> Hello @mwithi ! I've tested with the Arnaud's backend. Dates are showing correctly. Arnaud just need to perform check for the filter that's returning no content.

with the latest commit the date is showing, but not the disease

image

mwithi (approved):

PR #503, opened 2023-08-03
OH2-215 | Update styles for print media

SteveGT96:
See OH2-215.


mwithi:
Hello @SteveGT96, the print format looks quite fine but it totally miss any info about patient's personal data (left widget)
Is it possible to inlcude a box at the top of the page with at least First Name, Second Name, Patient ID and the rest of "Personal Data" section?

mwithi:
> Hello @SteveGT96, the print format looks quite fine but it totally miss any info about patient's personal data (left widget). Is it possible to inlcude a box at the top of the page with at least First Name, Second Name, Patient ID and the rest of "Personal Data" section?

Also, how this solution will behave using "print" on any other page?

SteveGT96:
> Hello @SteveGT96, the print format looks quite fine but it totally miss any info about patient's personal data (left widget). Is it possible to inlcude a box at the top of the page with at least First Name, Second Name, Patient ID and the rest of "Personal Data" section?

Let me check it !

mwithi:
> Let me check it !

Looks nice!

mwithi:
> Also, how this solution will behave using "print" on any other page?

What about this? For example when doing CTRL+P on other pages the result is not good, seems specific for Clinic page only, right?

e.g. for Admissions page:

image

SteveGT96:
It's strange ! Here's how I'm getting the admission page.
How do you find it ?
Screenshot from 2023-08-16 17-07-16

mwithi:
> It's strange ! Here's how I'm getting the admission page. How do you find it ?

1. I'm using demo data
2. If you refer to the grey background then it's just the printing option for "graphics on background"
3. I refert to the fact that the warning message should not be there, but in general I'm asking if there is a way to customize the "print mode" per page or it is general for all pages

SteveGT96:
> > It's strange ! Here's how I'm getting the admission page. How do you find it ?
> 
> 1. I'm using demo data
> 2. If you refer to the grey background then it's just the printing option for "graphics on background"
> 3. I refer to the fact that the warning message should not be there, but in general I'm asking if there is a way to customize the "print mode" per page or it is general for all pages

Hello @mwithi . In the last commit I've pushed, I've removed info box on print mode. The customization is general for all pages(but the patient infos are just for the pages inside the patient details activity). You could specify comfiguration per page and I'll try to fix that up.

mwithi:
> Hello @mwithi . In the last commit I've pushed, I've removed info box on print mode. The customization is general for all pages(but the patient infos are just for the pages inside the patient details activity). You could specify comfiguration per page and I'll try to fix that up.

with the latest commit looks fine for me, we can merge this PR as for now. thanks!

mwithi (approved):
LGTM. Thanks a lot!

PR #507, opened 2023-08-08
OH2-222 | Align with they latest api spec file and add ward in visit form and table

SteveGT96:
See OH2-222.


mwithi (changes requested):
It seems ok, but please remove also the "+ CREATE NEW OPD VISIT” button

mwithi:
Up!

mwithi:
I think header is ok (we should think about how to change the logo though)

image

pay attention that long notes are not wrapped and scrollbars should be avoided, as well as untranslated labels (see _angal.lab.blood.txt_)

image

mwithi:
Also, seems there are conflicts with `src/styles/index.scss`

src/generated/runtime.ts
mwithi, line 1: `withCredentials` has to stay

src/generated/apis/LoginApiApi.ts
mwithi, line 1: why so many files are deleted?

SteveGT96: When I was working on the task, I replaced the api file with the one provided by the backend.

PR #509, opened 2023-08-31
OH2-215 | Add toggle expand table component data shorcut(ctrl+space)

SteveGT96:
When printing data containing table with expandable content, to avoid content being partially shown in print mode, we could use Ctrl+Space to expand all expandable table row before using Ctrl+P. We could also use Ctrl+Space to toggle expand/collapse all.


mwithi:
Using CTRL+Space before CTRL+P works, but it is not an out-of-the-box solution for the end user, who might never know about this procedure. Is it possible to trigger the CTRL+Space automatically when the user wants to print?

mwithi (approved):
LGTM. Thanks a lot!

PR #518, opened 2023-09-18
OH2-228 | Add hospital info on print

SteveGT96:
See OH2-228.

This add the hospital info on the top of some pages when printing.

For the moment, Hospital infos have been added on the top of pages located in the activities PatientDetails, Laboratory, Visit. 
They could be added in other pages if necessary, using the HospitalInfo component.

### DIFF
--- a/src/components/accessories/hospitalInfo/HospitalInfo.tsx
+++ b/src/components/accessories/hospitalInfo/HospitalInfo.tsx
@@ -1,6 +1,5 @@
 import React, { FC, useEffect } from "react";
 import { useTranslation } from "react-i18next";
-import { useNavigate } from "react-router";
 import logo from "../../../assets/logo-color.svg";
 import "./styles.scss";
 import { TProps } from "./types";
--- a/src/components/accessories/hospitalInfo/HospitalInfo.tsx
+++ b/src/components/accessories/hospitalInfo/HospitalInfo.tsx
@@ -1,0 +1,43 @@
+import React, { FC, useEffect } from "react";
+import { useTranslation } from "react-i18next";
+import { useNavigate } from "react-router";
+import logo from "../../../assets/logo-color.svg";
+import "./styles.scss";
+import { TProps } from "./types";
+import { IState } from "../../../types";
+import { useDispatch, useSelector } from "react-redux";
+import { getHospital } from "../../../state/hospital/actions";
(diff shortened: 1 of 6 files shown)


mwithi:
I think header is ok (we should think about how to change the logo though)

image

pay attention that long notes are not wrapped and scrollbars should be avoided, as well as untranslated labels (see angal.lab.blood.txt)

image

Also, seems there are conflicts with src/styles/index.scss

SteveGT96:
Hello @mwithi ! For labels translation, I think we'll need a standard to manage internationalized fields. For the unwrapped content, I'm not able to reproduce it.

mwithi:
> Hello @mwithi ! For labels translation, I think we'll need a standard to manage internationalized fields. For the unwrapped content, I'm not able to reproduce it.

The LabManager in core returns an HashMap with "keys" -> translations, you can you use those "keys" as any other language key for UI translations.
Anyway, the material can be hidden in the UI, and so its translations, see https://openhospital.atlassian.net/browse/OHCS-73

SteveGT96:
> > Hello @mwithi ! For labels translation, I think we'll need a standard to manage internationalized fields. For the unwrapped content, I'm not able to reproduce it.
> 
> The LabManager in core returns an HashMap with "keys" -> translations, you can you use those "keys" as any other language key for UI translations. Anyway, the material can be hidden in the UI, and so its translations, see https://openhospital.atlassian.net/browse/OHCS-73

Is there yet an endpoint that serves those translations ?

mwithi:
> Is there yet an endpoint that serves those translations ?

`/laboratories/materials`

SteveGT96:
OK ! I think if there is not matter with the Hospital Header  info then we could merge this PR and then create issues for extra minor bugs.

mwithi:
> OK ! I think if there is not matter with the Hospital Header info then we could merge this PR and then create issues for extra minor bugs.

Ok, but the issue for the material is already in progress (to hide them), I will open the one for the hospital logo

mwithi (approved):
LGTM. Thanks a lot!

SteveGT96:
> > OK ! I think if there is not matter with the Hospital Header info then we could merge this PR and then create issues for extra minor bugs.
> 
> Ok, but the issue for the material is already in progress (to hide them), I will open the one for the hospital logo

@SilverD3 is working on it !

PR #526, opened 2023-10-20
OH2-245 | Update triage default values

SteveGT96:
See OH2-245.

Paired with https://github.com/informatici/openhospital-core/pull/1136 and https://github.com/informatici/openhospital-api/pull/371


mwithi (changes requested):
First -> we have to work on issues of type OP or OH2 (potentially linked to one OHCS) and then the PRs should refer to first them.

Second -> this solution is forcing the three values to their "normal" values but it should be allowed `null` values as well, which means the sign has not been evaluated by the doctor (there is indeed a difference between "normal" observation and not observed).

Same applies for all other values, with the exception for `height` and `weight`: 1. they should be compulsory and 2. in the previous GUI their values (only height weight) are taken from the last triage, if any, in order to help the doctor.

SteveGT96:
> First -> we have to work on issues of type OP or OH2 (potentially linked to one OHCS) and then the PRs should refer to first them.
> 
> Second -> this solution is forcing the three values to their "normal" values but it should be allowed `null` values as well, which means the sign has not been evaluated by the doctor (there is indeed a difference between "normal" observation and not observed).
> 
> Same applies for all other values, with the exception for `height` and `weight`: 1. they should be compulsory and 2. in the previous GUI their values (only height weight) are taken from the last triage, if any, in order to help the doctor.

We need then endpoint to provide the last triage and also the current endpoint to handle null or empty values. I've set default values cause the backend was rejecting null values.

What do you thing ?

mwithi:
> We need then endpoint to provide the last triage and also the current endpoint to handle null or empty values. I've set default values cause the backend was rejecting null values.
> 
> What do you thing ?

`null` values will be accepted when merged https://github.com/informatici/openhospital-api/pull/371

endpoint for last triage should be `/examinations/lastByPatientId/{patId}`

SteveGT96:
> > We need then endpoint to provide the last triage and also the current endpoint to handle null or empty values. I've set default values cause the backend was rejecting null values.
> > What do you thing ?
> 
> `null` values will be accepted when merged [informatici/openhospital-api#371](https://github.com/informatici/openhospital-api/pull/371)
> 
> endpoint for last triage should be `/examinations/lastByPatientId/{patId}`

OK Perfect !

mwithi:
Superseded by https://github.com/informatici/openhospital-ui/pull/528

PR #535, opened 2023-12-01
OH2-193 | Display dashboard components base on user role/permission

SilverD3:
See OH2-193.

### Changes 
- Align APIs and Models with spec file.
- Update `TPermission` type
- Use Authenticated user permission to display dashboard widgets


SilverD3:
The API `PUT /users/settings/{1}` used to update UserSetting is not working. That's why I marked this PR as draft.

SilverD3:
Something like "A setting with the same name already exists"

mwithi:
> Something like "A setting with the same name already exists"

pushed fix on api, please check, thanks

SilverD3:
Looks good. 

I've test with the four users in the demo database.

mwithi (approved):
LGTM. Thanks a lot!

PR #538, opened 2023-12-13
OH2-255 Revise and align permissions for CORE, API and UI

mwithi:
See OH2-255.

- [x] Fix UI with redefined permissions in CORE and API
- [x] move `/discharges` to `/admissions/discharges` endpoint and remove related permission (admission.* ones will be used)
- [x] change `visits.access` to `opds.access` (visits is another Entity with other purposes, see OH2-256)
- [x] fix `exam.*` permissions to `laboratories.*` when doing Laboratory activities (see comments)


almeida1492 (approved):
lgtm

PR #539, opened 2023-12-19
OH2-259 | Fix error message on Laboratory page

SteveGT96:
See OH2-259.


src/resources/i18n/en.json
```diff
@@ -562,10 +562,10 @@
     "changestatus": "Change Exam Status",
     "changelabstatusto": "The status of the exam {{code}} will be changed to {{status}}",
     "statuses": {
-      "DRAFT": "DRAFT",
-      "ALL": "ALL",
-      "OPEN": "OPEN",
-      "DONE": "DONE"
+      "dragt": "DRAFT",
```
mwithi, line 565:
```suggestion
      "draft": "DRAFT",
```

src/resources/i18n/de.json
mwithi, line 1: there's no need to provide translations other than "en" because they will be automatically handled by Transifex

src/resources/i18n/en.json
```diff
@@ -4,7 +4,10 @@
     "incorrectcredentials": "Invalid username or password",
     "passwordtooshort": "Password too short",
     "badrequest": "Bad request",
-    "internalerror": "An internal error occured"
+    "internalerror": "An internal error occured",
+    "lab": {
+      "someexamswithoutresultpleasecheck": "Some exams without result please check !"
```
mwithi, line 9:
```suggestion
      "someexamswithoutresultpleasecheck": "Some exams without result please check."
```

mwithi: and probably in ExamForm we need `<InfoBox type="error" message={t(errorMessage)} />`

mwithi: There's no need of esclamation point, see my suggestions please

mwithi (approved):
LGTM. Thanks a lot!

PR #544, opened 2024-01-17
OH2-262 | Update admission ward field validation

SteveGT96:
See OH2-262.

### DIFF
--- a/src/components/accessories/admission/admissionForm/AdmissionForm.tsx
+++ b/src/components/accessories/admission/admissionForm/AdmissionForm.tsx
@@ -57,7 +57,9 @@ const AdmissionForm: FC<AdmissionProps> = ({
   const admissionTypes = useSelector(
     (state: IState) => state.admissionTypes.allAdmissionTypes.data
   );
-  const wards = useSelector((state: IState) => state.wards.allWards.data);
+  const wards = useSelector((state: IState) =>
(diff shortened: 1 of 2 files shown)


mwithi (changes requested):
I think the ward should not been even shown, because imagine if you a long list of wards mixed for IPD and OPD... it is better to show only the OPD ones

Also, I think that "Date" should be "Admission Date" and I don't know why it complains about the Discharge Date that is not even there in the form while we are starting an admission. Please could you check also this?

image

src/components/accessories/admission/admissionForm/AdmissionForm.tsx
```diff
@@ -57,7 +57,9 @@ const AdmissionForm: FC<AdmissionProps> = ({
   const admissionTypes = useSelector(
     (state: IState) => state.admissionTypes.allAdmissionTypes.data
   );
-  const wards = useSelector((state: IState) => state.wards.allWards.data);
+  const wards = useSelector((state: IState) =>
+    state.wards.allWards.data?.filter((ward) => !(ward.opd ?? false))
```
mwithi, line 61: as the issues states, the wards to be filtered are the ones with 0 `beds`, not the ones with `opd=true` because that it only tell us if the ward is accepting OPD visits.

mwithi (approved):

PR #546, opened 2024-01-17
OH2-264 | Sort table component date fields by date time

SteveGT96:
See OH2-264.


mwithi (changes requested):
Something wrong (blank page) happens when I try to see Clinics with more than one record to sort.

image

mwithi:
Ok, records looks correctly ordered now, can we have the "time" information somewhere please? I think that in the record is ok only the date, but in the expanded part it would be bettere to have date and time. 

image

For the visits the date field is missing in the expanded part.

Also, the header in "Previous Visits" looks different ("Date DD/MM/YY")

image

src/libraries/sortUtils/sortUtils.ts
```diff
@@ -6,6 +6,8 @@ const formats = [
   "YYYY-MM-DD",
   "YYYY-MM-DD HH:mm:ss",
   "DD/MM/YYYY HH:mm:ss",
+  "YYYY-MM-DD HH:mm",
+  "DD/MM/YYYY HH:mm",
```
gasp, line 10: This is a strange non-standard format (it looks like the [French date notation format](https://en.wikipedia.org/wiki/Date_and_time_notation_in_France)). Shouldn't we avoid non-standard formatting in data?

mwithi: I think only "DD/MM/YYYY HH:mm" is enough as international way of understanding dates and time, although in the future we may consider different "users' settings" for rendering

src/generated/runtime.ts
```diff
@@ -15,7 +15,7 @@ import { Observable, of, Subscriber } from "rxjs";
 import { ajax, AjaxRequest, AjaxResponse } from "rxjs/ajax";
 import { map, concatMap } from "rxjs/operators";
 
-export const BASE_PATH = "https://oh2.open-hospital.org".replace(/\/+$/, "");
+export const BASE_PATH = "http://localhost:8080".replace(/\/+$/, "");
```
mwithi, line 18: Please revert this change

mwithi (approved):

PR #551, opened 2024-01-25
chore: restrict billing utils to billing

gasp:
moving utils functions `combineData` and `sortAndSlice` into the billing folder as :
* their expected behaviour is not obvious
* they are only used in billsMining.ts

I also added tests up to my understanding of their behavious

### DIFF
--- a/src/components/accessories/billsRecap/billsMining.test.ts
+++ b/src/components/accessories/billsRecap/billsMining.test.ts
@@ -1,0 +1,36 @@
+import { combineData, sortAndSlice } from "./billsMining";
+
+describe("billsMining", () => {
+  describe("combineData", () => {
+    it("should correctly combine data", () => {
+      const input = { key1: "value1", key2: "value2" };
+      const output = combineData(input);
+
+      expect(output).toEqual(input);
+    });
+  });
+
+  describe("sortAndSlice", () => {
+    it("should sort", () => {
+      const input = { keyA: 1, keyB: 2, keyC: 3, keyD: 4 };
+      const output = sortAndSlice(input);
+      expect(output).toEqual({ keyD: 4, keyC: 3, keyB: 2, keyA: 1 });
+    });
+    it("should sort", () => {
+      const input = { keyA: 1, keyB: 1, keyC: 1, keyD: 3 };
+      const output = sortAndSlice(input);
+      expect(output).toEqual({ keyA: 1, keyB: 1, keyC: 1, keyD: 3 });
+    });
+    it("should slice", () => {
+      const input = Array.from({ length: 99 }, (_v, index) => index).reduce(
+        (acc, index) => {
+          acc[`key${index}`] = index;
+          return acc;
+        },
+        {} as Record<string, number>
+      );
+      const output = sortAndSlice(input);
+      expect(Object.keys(output).length).toBe(10);
+    });
+  });
+});
--- a/src/components/accessories/billsRecap/billsMining.ts
+++ b/src/components/accessories/billsRecap/billsMining.ts
@@ -1,10 +1,24 @@
 import moment from "moment";
 import { FullBillDTO } from "../../../generated";
-import {
-  combineData,
-  sortAndSlice,
-} from "../../../libraries/formatUtils/dataFormatting";
 import { IBillSummary } from "../../activities/billingActivity/types";
+
+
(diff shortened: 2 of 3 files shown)


SteveGT96:
Hello @gasp ! In MHO, sortAndSlice and combineData could be used in another places out of the scope of bills. It could be therefore correct to keep them in the libs(libraries) folder; also their unit tests. For sortAndSlice, did you test also the sort ?
Thanks.

mwithi:
Maybe we can keep those functions in `formatUtils` and add some comments to better clarify them?

gasp:
Hey all, thanks for your comments and taking time to code review.
As I am new to the project, I didn't want to make any change that could affect the business logic. Also, I still struggle with the whole billing part.

I wanted to isolate those two functions because they seem (to me, and I might be wrong) prone to errors.
Isolating those functions that are only used by the `computeBillSummary` function would avoid that they spread to other parts of the code and introduces unexpected behaviours (or bugs).

## combineData:

As you can see in [this test](https://github.com/informatici/openhospital-ui/pull/551/files#diff-9b044f4da80d21b5ea830a6c013e0410c0c01bd7ff49358b4ca737febcc432b9R9), the function essentially creates a shallow copy of the input object. If the values in the input object are objects or arrays, they will be referenced in the new object. This is an _unorthodox_ way of coding in javascript. It may introduce confusion and I still don't get what purpose it serves.

It looks like it _normalizes_ various objects (arrays, strings…) [into an object](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Object/entries).

Can it be refactored as `const CombineData = (obj: any): Record<string, any> => {...obj};` ?.
Anyway, the [any:any typing](https://github.com/informatici/openhospital-ui/pull/551/files#diff-9816499fcb87d61c827aa9159e6e7a5cb930559eee2829c6ebf07ce3f92f87adL47) is more than ambiguous and should be refactored.

Also, it does not "combine" any data, it reformats sth as an object, so why is it called combineData 🤔?


## sortAndSlice

Same thing here, the any:any typing makes me uncomfortable.

And by adding tests, we demonstrate that it is not really sorting: 
* [here](https://github.com/informatici/openhospital-ui/pull/551/files#diff-9b044f4da80d21b5ea830a6c013e0410c0c01bd7ff49358b4ca737febcc432b9R17) it sorts values from bigger to lower
* [here](https://github.com/informatici/openhospital-ui/pull/551/files#diff-9b044f4da80d21b5ea830a6c013e0410c0c01bd7ff49358b4ca737febcc432b9R22) it sorts values from lower to bigger

This looks like a bug and should need further investigations.

Also, [this line](https://github.com/informatici/openhospital-ui/pull/551/files#diff-2ad5dd22d6b8cb490b52a84bd834d70d5f1ffc78e592bf783d421e164cdb3e2fR20) does the same thing as combineData, and could be replaced by combineData because it's the same code just formatted another way. So we are doing the same strange shallow referencing.

---

Thanks for taking the time reading all this. AgainI have little knowledge on this project and please excuse me if I am terribly missing the point.

mwithi:
thanks @gasp for the careful analysis, I think we should definetely look into it, wdyt @SteveGT96 and @SilverD3 ?

SteveGT96:
@gasp's remarks are interesting. For the moment we don't have billing in the current MVP(am I wrong?), I think therefore that we could look deeper that at the correct moment. Meantime, @gasp could take a look at what we've been focused on until now. Wdyt @mwithi ?

mwithi:
hello @SteveGT96, I don't think @gasp is awared about MVP and latest Sprint (unless he has read the latest meetings notes and the development ML 🙂), so I think that his advice might be not urgent but still very valuable.

@gasp do you think you can provide a better version for the two methods in this PR (rather than just moving them) provided that you don't need further information of what those two methods should do (or not do)?

gasp:
You are right, it is not the proper way to address this issue, I should dive deeper into this feature and fix it instead of just scratching the surface.

I am experimenting issues testing it properly, and understanding the expectations, I'll work on it when it is a sprint goal. Sorry for taking your time on this it's diluting our focus.

mwithi:
Thank you @gasp, if you want to join the discussions and the online meetings anytime (I give notice on Slack and ML-devel) you're welcome!!

PR #557, opened 2024-02-09
OH2-269 | Fix vulnerability 'Inefficient Regular Expression Complexity in nth-check'

SteveGT96:
See OH2-269.

Before (on develop branch)
npm audit -> 14 vulnerabilities (4 moderate, 10 high)
npm audit --production -> 10 vulnerabilities (4 moderate, 6 high)

Now (On this PR)
npm audit -> 13 vulnerabilities (4 moderate, 10 high)
npm audit --omit=dev -> 0 vulnerabilities


SilverD3 (approved):
The vulnerability is fixed in prod, but not in dev. As we are first fixing prod vulnerability, I think we can merge the PR

mwithi:
> The vulnerability is fixed in prod, but not in dev. As we are first fixing prod vulnerability, I think we can merge the PR

what do you mean with prod and dev? what is "prod"? 
Anyway there are conflicts to be resolved @SteveGT96

SilverD3:
'prod' is production and 'dev' is development. Some vulnerabilities only affects the development environment as the related deps are only used during development.
`nth-check` is related to a dependency that @SteveGT96 moved to dev dependencies, so when the project is audited using production configuration, the vulnerability being fixed in this PR is not longer shown

SilverD3:
> Ok, what I'm asking is how you differentiate between dev and prod configuration? Will https://github.com/dependabot always check both so the security warning will be shown over time?

In package.json, there is a property (`devDependencies`) to specify dependencies that we only use in development mode, like a mock server lib. When running npm audit, you can specify if you're auditing the app in production mode (it will not inspect dev deps) with option `--prod` or just auditing all deps

mwithi:
> I think @dependabot only checks vulnerabilities for production, but I need to perform further research to clarify it

Ok, thanks a lot for all explanations and let us know!

mwithi (approved):
LGTM. Thanks a lot!

PR #568, opened 2024-02-21
fix: i18n status in patientActivity menu

gasp:
Added internationalisation for statuses in patientActivity

before / after

beforeafter


mwithi (approved):
LGTM. Thanks a lot!

PR #589, opened 2024-05-23
fix(fixtures): make fixtures more expresive / easier to test with

gasp:
this enables to better test diseases and diseases types.

Before:

before

After:

after


SilverD3 (approved):

mwithi (approved):
Approved by @SteveGT96

PR #590, opened 2024-05-23
fix: prevent from crashing when patientContext isn't resolved yet

gasp:
It happened while I was testing something else. In all other contexts, this is safely handled, but not here. So I am fixing it.

Other ocurences in the code where this exception is handled
* https://github.com/informatici/openhospital-ui/blob/develop/src/components/accessories/admission/PatientAdmission.tsx#L199
* https://github.com/informatici/openhospital-ui/blob/develop/src/components/activities/patientDetailsActivity/PatientDetailsActivity.tsx#L261
* https://github.com/informatici/openhospital-ui/blob/develop/src/components/activities/searchPatientActivity/PatientSearchItem.tsx#L56


SteveGT96 (approved):
Great @gasp !

mwithi (approved):
Approved by @SteveGT96

PR #586, opened 2024-05-22
OH2-291 | Vaccine list

SilverD3:
See [OH2-291](https://openhospital.atlassian.net/browse/OH2-291)

[OH2-291]: https://openhospital.atlassian.net/browse/OH2-291?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

PR #591, opened 2024-05-24
OH2-302 | OH2-296 | Types main layout and vaccine types CRUD

SilverD3:
See OH2-302 and OH2-296


src/mockServer/server.js
```diff
@@ -1,62 +1,64 @@
-import XHRAdapter from "@pollyjs/adapter-xhr";
```
gasp: What's with this change?

Why is git diff catching the whole file as a change? Did you change [file encoding](https://en.wikipedia.org/wiki/Character_encoding) or [line-ending](https://en.wikipedia.org/wiki/Newline) a.k.a CRLF?

(or am I worrying too much 😅?)

gasp:
This is really great stuff, thanks Sylvester!

I checked the code and everything looks nice and clean. Unfortunately I haven't had the time yet to run it on my machine and test it myself. @SteveGT96 could you test on your machine? The more feedback, the better 😀

SteveGT96 (approved):
AFAIK, looks OK for me(when @gasp  comment will be resolved) ! But concerning the css styles, we should also consider @axelpezzo's review. @mwithi please, could you request @axelpezzo's review ?

mwithi:
> AFAIK, looks OK for me(when @gasp comment will be resolved) ! But concerning the css styles, we should also consider @axelpezzo's review. @mwithi please, could you request @axelpezzo's review ?

Done. And also, @SilverD3 could you process the @gasp comment?

mwithi:
@SilverD3 please resolve conflicts also (I think it is the topic suggested by @gasp on [Slack](https://join.slack.com/t/openhospitalworkspace/shared_invite/zt-cg9iu1d0-sjTuWWziFgcDqiTWyMJBAw))

src/mockServer/server.js
SilverD3: I don't know why git is considering the whole file content has changed. I didn't change file encoding or line-ending. So if there's no problem related to this behavior, we're good to go.

axelpezzo:
I think that we should standardize the style in due to the fact that now there are some misleading inside the components.

We should use the .scss file rather than line-styling or tailwind.

src/components/accessories/admin/types/styles.scss
```diff
@@ -0,0 +1,27 @@
+@import "../../../../styles/variables";
+@import "../../../../../node_modules/susy/sass/susy";
```
gasp, line 2: As [from offical doc](https://github.com/oddbird/susy): "Susy is Deprecated. It should no longer be used on new projects, and will no longer be receiving any updates." should we stop using it?

Yet, there are [50 other occurences of susy-media](https://github.com/search?q=repo%3Ainformatici%2Fopenhospital-ui+susy-media&type=code), so this is totally opened to discussion and if this is a technical change we want to do, let's handle it separately. wdyt?

src/components/activities/adminActivity/AdminActivity.module.scss
```diff
@@ -14,7 +14,7 @@
   border-radius: 8px;
   flex-grow: 1;
   margin: 30px 160px;
-  align-self: center;
+  // align-self: center;
```
gasp: master-yoda-star-wars

SilverD3: I commented out this rule because it's causing the layout to shrink or grow depending on the content. I didn't remove it completely to keep a record, since I wasn't the one who initially defined this rule.

SilverD3:
> I think that we should standardize the style in due to the fact that now there are some misleading inside the components.
> 
> We should use the .scss file rather than line-styling or tailwind.

We should define utilities CSS classes that can be used accros the project. For the moment when I want to add margins, padding, or use grid or flex boxes in a component, I have to create CSS rules from scratch. And later, if a need to do it again in another component, I recreate from scratch again, and so on. 

The idea of ​​creating a .scss file in which there will be a single class with a single CSS rule and that this class will only be used once seems kind of wasteful to me. But if we need to proceed like this for a standardization, it's okay.

src/components/activities/adminActivity/AdminActivity.module.scss
gasp: Please don't worry about keeping a record, `git blame` can be used if necessary.
Please don't hesitate, [be bold](https://en.wikipedia.org/wiki/Wikipedia:Be_bold). It's open source software, if you see something wrong, please fix it!

If you think that this rule is incorrect, please remove it.
If you think that it's correct, please leave it.

If you make a change, test.
If you doubt, and don't understand why it was there before, please test even more :-)

mwithi:
While waiting @axelpezzo review, let me report that under vaccines I see wards
and for vaccinetypes looks something wrong with permission (admin user have all permissions)

chrome-capture-2024-6-7

chrome-capture-2024-6-7 (1)

SilverD3:
@mwithi About the first picture, the feature is provided by #586 which is blocked by this PR
For what concerns the second picture, as you can see, the user you're using doesn't have permissions to get vaccine types. This issue is related to backend because with the mock data, everything runs smoothly. I reported the issue to Arnaud

mwithi:
> @mwithi About the first picture, the feature is provided by #586 which is blocked by this PR For what concerns the second picture, as you can see, the user you're using doesn't have permissions to get vaccine types. This issue is related to backend because with the mock data, everything runs smoothly. I reported the issue to Arnaud

As I said I'm using admin user, that has all permissions

image

Anyway, let's merge and I will test again

mwithi (approved):
Approved by @gasp and @SteveGT96

PR #561, opened 2024-02-14
OH2-270 | Fix chalk/ansi-regex and shelljs vulnerabilities

SteveGT96:
**This PR include:**

- Cypress migration from 6 to 13 + specs migration to typescript.

- Storybook migration from 6 to 7.

Vulnerabilities related to chalk/ansi-regex and shelljs have been fixed(priority to production environment). The Upgrade of cypress and storybook introduces may other vulnerabilities fix like shown bellow.

**Before (on develop branch)**
npm audit -> 53 vulnerabilities (19 moderate, 34 high)
npm audit --production -> 18 vulnerabilities (7 moderate, 11 high)

**Now (On this PR)**
npm audit -> 14 vulnerabilities (4 moderate, 10 high)
npm audit --production -> 10 vulnerabilities (4 moderate, 6 high)

PR #592, opened 2024-05-26
ci: try to fix ci e2e

gasp:
Since https://github.com/informatici/openhospital-ui/pull/561 it seems that ci e2e jobs are failing.

error message says ([source](https://github.com/informatici/openhospital-ui/actions/runs/9210777728/job/25338565713#step:3:776))

```
You are attempting to use Cypress with an older config file: cypress.json
When you upgraded to Cypress v10.0 the config file was updated and moved to a new location: cypress.config.ts

You may need to update any CLI scripts to ensure that they are referring the new version. This would typically look something like:
"cypress open --config-file=cypress.config.ts"
```

we are trying to fix that

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -41,3 +41,4 @@ jobs:
           record: false
           start: npm start
           wait-on: "http://localhost:3000"
+          config-file: cypress.config.ts


gasp:
My bad, this is not the issue.

By upgrading from `cypress-io/github-action@v2` to `cypress-io/github-action@v6` this shall be automatically handled. No need to be verbose here, just make sure we are running the latest version of cypress-io/github-action :-)


This only fails on develop branch, it's because `ci.yml` and `deploy.yml` aren't aligned ([source](https://github.com/informatici/openhospital-ui/blob/develop/.github/workflows/deploy.yml#L38))

Please see upcoming pull request :-)

PR #597, opened 2024-05-30
OH2-307 | Refactor api response type

SteveGT96:
This replace the IApiResponse interface with the ApiResponse class. Previously hasSucceeded and isLoading were read/write properties in defined types.
With this refactoring, isLoading and hasSucceeded are just getters and their returned value relies on the status proprerty.

There are two additional getters: hasFailed, isSuccessEmpty that rely also on the status property.

See [OH2-307](https://openhospital.atlassian.net/browse/OH2-307)

[OH2-307]: https://openhospital.atlassian.net/browse/OH2-307?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/src/components/accessories/admin/diseases/diseaseTable/DiseaseTable.tsx
+++ b/src/components/accessories/admin/diseases/diseaseTable/DiseaseTable.tsx
@@ -6,7 +6,7 @@ import { CircularProgress } from "@material-ui/core";
 import { useSelector } from "react-redux";
 import { IState } from "../../../../../types";
 import { DiseaseDTO } from "../../../../../generated";
-import { IApiResponse } from "../../../../../state/types";
+import { ApiResponse } from "../../../../../state/types";
 import classes from "./DiseaseTable.module.scss";
 import { CheckOutlined } from "@material-ui/icons";
 import { TFilterField } from "../../../table/filter/types";
@@ -67,7 +67,7 @@ export const DiseaseTable: FunctionComponent<IOwnProps> = ({ onEdit }) => {
 
   const { data, status, error } = useSelector<
     IState,
-    IApiResponse<DiseaseDTO[]>
+    ApiResponse<DiseaseDTO[]>
   >((state) => state.diseases.allDiseases);
 
   const handleEdit = (row: DiseaseDTO) => {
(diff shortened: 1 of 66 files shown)


gasp:
Thanks for the pull request, I appreciate the effort of explaining your point of view supported by a solid implementation. It states very clearly what you want to do, how you want to implement it and how it is impacting the codebase.


I feel that this is adding complexity as it is giving 2 different ways to do the same thing.

```
if (something.status === "LOADING") {
```
and

```
if (something.isLoading) {
```

Concerning maintainability, it is always better to give only one obvious way to handle things and avoid "magic automated behaviors". Why not relying only on the status?

SteveGT96:
> Thanks for the pull request, I appreciate the effort of explaining your point of view supported by a solid implementation. It states very clearly what you want to do, how you want to implement it and how it is impacting the codebase.
> 
> I feel that this is adding complexity as it is giving 2 different ways to do the same thing.
> 
> ```
> if (something.status === "LOADING") {
> ```
> 
> and
> 
> ```
> if (something.isLoading) {
> ```
> 
> Concerning maintainability, it is always better to give only one obvious way to handle things and avoid "magic automated behaviors". Why not relying only on the status?

As you can see in your example,

(something.isLoading) is just a shorcut for (something.status === "LOADING")

`
get isLoading(): boolean {
    return this.status === "LOADING";
 }
`

You can just consider isLoading as an helper. If you like [state.status === "LOADING"], then no problem. However it doesn't hurt to write [state.isLoading].

gasp:
IMO this logic creates 2 different ways to do the same thing, it adds a little complexity without enhancing significantly the developer experience.
For this reason, I would prefer not to go that way.

But I am totally fine with other opinions and if other contributors prefer this way, I'll happily comply :-)

SilverD3:
@gasp I don't think @SteveGT96 is adding a level of complexity as the properties `hasSucceeded` and `isLoading` already exists in `ApiResponse`. He's actually simplifying things because when you're managing state, you have to update `isLoading`, `hasSucceeded` and `status` properties, but you'll agree with me that we can determine the value of `isLoading` and `hasSucceeded` based on `status`.

gasp:
Hi @SilverD3 thanks for taking the time to support your point. It is clear now why @SteveGT96 opened this pull request.
I'll align my code to fit this way of handling status in my own PRs.

gasp (approved):

mwithi (approved):
Approved by @gasp

PR #596, opened 2024-05-29
OH2-305 | Suppliers create and update

SilverD3:
The API doesn't provide any endpoint for suppliers deletion, so the delete feature isn't include in this PR.

See OH2-305

PR #599, opened 2024-06-05
OH2-309 | Minor bugs fix

SteveGT96:
See [OH2-309](https://openhospital.atlassian.net/browse/OH2-309)

[OH2-309]: https://openhospital.atlassian.net/browse/OH2-309?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

PR #600, opened 2024-06-06
fix: remove admin prefix from formatted urls

gasp:
Fixing an issue introduced in https://github.com/informatici/openhospital-ui/pull/596
as AdminRouter only applies in /admin/ let's remove this prefix when catching them.

this allows to catch `/admin/wards`, instead of current (wrongly) catching `/admin/admin/wards`

Screen Recording 2024-06-06 at 17 16 29


mwithi:
maybe duplicate of #599

gasp:
oopsie, thanks @stevegt96 for the fix!

PR #603, opened 2024-06-09
fix(oh2-314): userGroupNames is an object, not a string

gasp:
In the component, UserGroupName is displayed as a string while mocked, but it's an object [in the API](https://github.com/informatici/openhospital-api/blob/develop/openapi/oh.yaml). Which causes the ui to throw:
image (6)

Want to know more? Check the issue [OH2-314](https://openhospital.atlassian.net/browse/OH2-314)

Side note for later, to avoid further issues, fixtures should be typed against generated data from the API.

[OH2-314]: https://openhospital.atlassian.net/browse/OH2-314?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


mwithi (approved):
LGTM. Thanks a lot!

PR #605, opened 2024-06-10
OH2-295 | Types / Operation CRUD

SilverD3:
See OH2-295


mwithi (changes requested):
It performs a fake deletion, because OS is actually used and if I come back to the page the deleted record is still there

image
image
image

Note: after the deletion the filter shows only "Type of operations"

image

SilverD3:
> It performs a fake deletion, because OS is actually used and if I come back to the page the deleted record is still there

Are you using the mock server or the real backend? With the mock, it's normal the deleted type is reset after page/component refresh. But if you're using the real backend, that's where the issue would come from.

>Note: after the deletion the filter shows only "Type of operations"

Fixed!

mwithi:
it deletes

> > It performs a fake deletion, because OS is actually used and if I come back to the page the deleted record is still there
> 
> Are you using the mock server or the real backend? With the mock, it's normal the deleted type is reset after page/component refresh. But if you're using the real backend, that's where the issue would come from.

I always use the real backend.

> 
> > Note: after the deletion the filter shows only "Type of operations"
> 
> Fixed!

Thanks!

mwithi (approved):

PR #620, opened 2024-06-21
OH2-325 | Fix possibly bad designs

SilverD3:
See OH2-325

1. Before
We have to scroll to see action buttons
ward1
ward2

The error is displayed at the bottom of the page, causing an UX issue, as the user has to scroll to the bottom to see error message
ward-error-bottom

2. After
I've reduced the number of columns displayed by default to make the table compact and so make the action buttons visible without the need to scroll.
ward-min

I've also displayed the error message at the top of the page and an event to scroll to the message when it occurs
ward-error-top

I applied the fix in all admin functions

### DIFF
--- a/cypress/integrations/admin_activities/diseases_activities/manage_diseases_activity.cy.ts
+++ b/cypress/integrations/admin_activities/diseases_activities/manage_diseases_activity.cy.ts
@@ -13,7 +13,7 @@ describe("Diseases Activity specs", () => {
       .find("table")
       .then(($table) => {
         const rows = $table.find("tbody tr");
-        expect(rows.length).equal(3);
+        expect(rows.length).equal(6);
       });
   });
 });
--- a/src/components/accessories/admin/diseases/diseaseTable/DiseaseTable.tsx
+++ b/src/components/accessories/admin/diseases/diseaseTable/DiseaseTable.tsx
@@ -8,7 +8,7 @@ import { IState } from "../../../../../types";
 import { DiseaseDTO } from "../../../../../generated";
 import { ApiResponse } from "../../../../../state/types";
 import classes from "./DiseaseTable.module.scss";
-import { CheckOutlined } from "@material-ui/icons";
+import { CheckOutlined, CloseOutlined } from "@material-ui/icons";
 import { TFilterField } from "../../../table/filter/types";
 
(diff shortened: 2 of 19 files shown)


gasp:
in such cases, adding a small gif or image would be great to help the reviewer visualize the results

mwithi:
I think the point here is also that no matter the number of columns, if the window is squeezed enough the scrollbar will appear anyway

AND

if the table has too many rows, it is difficult to see both the scroolbar (unless reaching the end of the table) or the potential error message, regardless if it is put at the top or at the end of the page (unless on scrolls it all the way up or down)

So my idea behind OH2-325 was to think about how to fix the "design" not a specific page, such as the Ward one.

To have an idea please see Operation or Disases page (that this PR leave them unchanged at the moment...).

SilverD3:
Not sure to get you. The issue you mention not seems to be related to a design issue.
Please can you make a short video in which you explain the behavior (or maybe screenshots)?

mwithi:
When I say "design issue" I mean that maybe we should rethink the way it is designed, because despite it is working it is not user friendly.

The scrollbar is shown as soon the window is too narrow:

chrome-capture-2024-6-25

On long lists, the user doesn't know what is happening:

chrome-capture-2024-6-25 (1)

Anyway I would expect that changes in OH2-235 are reflected consistently in all pages.

SilverD3:
The issue OH2-235 you're pointing out is intended to display languages in their original spelling. I can't see how it could help to solve design issues

mwithi:
> The issue [OH2-235](https://openhospital.atlassian.net/browse/OH2-235) you're pointing out is intended to display languages in their original spelling. I can't see how it could help to solve design issues

Of course I meant OH2-325....

[OH2-235]: https://openhospital.atlassian.net/browse/OH2-235?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

mwithi (approved):
Although it solves only part of the issue, I'm going to merge because the initial issue had a goal too broad (my bad) so we are going to have dedicated issues for every fix

PR #621, opened 2024-06-24
feat(oh2-299): new user form

gasp:
[OH2-299](https://openhospital.atlassian.net/browse/OH2-299)
newuser

[OH2-299]: https://openhospital.atlassian.net/browse/OH2-299?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


SteveGT96 (changes requested):
Please, could you also add feedback dialog wether the operation succeeds or not ? (just like you've done for the Exam Form).

gasp:
Thanks for the review, here is how it looks like.

created

SilverD3:
LGTM!

In the [mock](https://w6nt6m.axshare.com/?code=373a8db2aee9ac436955a60fb179b091&id=kjyie9&p=add_user&g=1), we have `description` field, but I can see it in the form you provided.

gasp:
Here is the user description, sorry for omitting the field!

Screen Recording 2024-07-15 at 15 13 47

SilverD3 (approved):
It's working and tests are passing!

mwithi:
shouldn't the password be retyped twice?

mwithi:
Also, the group field shows the groups' descriptions but not the groups' names

image

We could show them as "name - description" I guess...

image

gasp:
> shouldn't the password be retyped twice?

Good idea, it shall reduce typing errors

gasp:
> shouldn't the password be retyped twice?
> Also, the group field shows the groups' descriptions but not the groups' names
> We could show them as "name - description" I guess.

* Password should be retyped twice
* group selector now displays "code - description"

Screen Recording 2024-07-16 at 14 00 10

gasp:
Small detail: autocomplete matches code or description
Screen Recording 2024-07-16 at 14 09 54

mwithi:
User creation is ok. If User already exists the error message is not defined (we need to improve api though)

image

mwithi:
I think that on the NEW USER page, we should also have a CANCEL button, and in case of error saving the USER (the message is ok now) it should not go back to the list but remain on the edit page

mwithi (approved):
Approved to move on with next developments

PR #624, opened 2024-07-08
OH2-335 | Manager enabled/disabled diseases

SteveGT96:
See OH2-335.


SilverD3 (commented):
I think you should add test cases for enabled/disabled diseases. It could be something like:
1. Go to diseases pages
2. Count rows to check if it matches some number
3. Click on "ENABLED"
4. Count rows to check if it matches some number
5. Click on "DISABLED"
4. Count rows to check if it matches some number

Also, I think there should be a space between "Expand all" and "diseases' states buttons"
Screenshot 2024-07-12 at 10-15-36 Open Hospital

SilverD3:
Good work so for. But it's not completely responsive

https://github.com/user-attachments/assets/a4d57e2b-5a94-4989-9a1f-ce506606ff25

SilverD3 (approved):

mwithi (approved):
LGTM. Thanks a lot!

PR #629, opened 2024-07-17
OH-332  | Add cypress e2e tests to cover delivery types (admin/types/de…

loique70:
See OH-332.


SilverD3 (changes requested):
Move `deliveries_activities` folder to `types_activities` folder

SilverD3 (approved):

PR #647, opened 2024-08-07
feat(oh2-298): admin edit user

gasp:
OH2-298

user-edit

this is ui only, it returns an error on current API, should be fixed by OH2-253


src/components/accessories/admin/users/editUser/EditUser.tsx
```diff
@@ -0,0 +1,64 @@
...
+
+export const EditUser = () => {
+  const dispatch = useAppDispatch();
+  const { id } = useParams();
+  const [user, setUser] = useState<UserDTO | null>();
+  const [userNotFound, setUserNotFound] = useState<boolean>(false);
+  const { isLoading, hasSucceeded, hasFailed, error } = useAppSelector(
+    (state) => state.users.update
+  );
+  const users = useAppSelector((state) => state.users.userList);
```
SilverD3: I think you should pass the user to edit in the router state or send a request to fetch the user to edit using his username (fetch only this user, not all users). 
Here you're fetching all users and this evolves two problems:
- If the endpoint to get all users is paginated, you might not find the user even though he exists.
- It is not the most efficient way to go about it.

src/components/accessories/admin/users/editUser/EditUser.tsx
```diff
@@ -0,0 +1,64 @@
...
+import { EditUserForm } from "./EditUserForm";
+
+export const EditUser = () => {
+  const dispatch = useAppDispatch();
+  const { id } = useParams();
+  const [user, setUser] = useState<UserDTO | null>();
+  const [userNotFound, setUserNotFound] = useState<boolean>(false);
+  const { isLoading, hasSucceeded, hasFailed, error } = useAppSelector(
+    (state) => state.users.update
+  );
```
SilverD3, line 24: You're not using these states vars in this component, but you're just passing to the child component. I think you should better fetch them in the child component directly from the store.

src/components/accessories/admin/users/usersTable/UsersTable.tsx
```diff
@@ -49,6 +50,7 @@ export const UsersTable = ({ headerActions }: IOwnProps) => {
         userGroupName:
           item.userGroupName?.desc ?? item.userGroupName?.code ?? "",
         desc: item.desc ?? "",
+        passwd: item.passwd ?? ""
```
SilverD3, line 68: Is the user password really sent (in plain text) to front-end (in the DTO) 😨 ?

SilverD3 (changes requested):

src/components/accessories/admin/users/usersTable/UsersTable.tsx
gasp: Yes, and it is stored in plain text in the database, which is not a best practice.
As the app is supposed to be used offline, I guess that's ok for now 😬

src/components/accessories/admin/users/editUser/EditUser.tsx
gasp: Passing the user in the state is not a good practice because it prevent from using the urls to navigate.
I am not confortable with fetching a single user, because it would put a wrong value in the state:

`state.users.userList` would then equal `{username: "john", password: "hello"}` or maybe `[{username: "john", password: "hello"}]`  which are both wrong because `state.users.userList` should be `null` or contain the actual user list. Abiding to this best practice opens the ability to do optimistic ui.

single user data should be stored in a format looking like: `state.users.user.john` where the state can be accessed directly by id.

More readings on this:
https://redux.js.org/usage/structuring-reducers/basic-reducer-structure
and https://redux.js.org/usage/structuring-reducers/normalizing-state-shape

wdyt?

cc @SteveGT96

src/components/accessories/admin/users/editUser/EditUser.tsx
gasp: That way I try to keep separated:
1. data concerns: querying from API, managing redux store
2. ui and local state concerns: formik, display of user groups, etc.

It is intended to provide a clearer and more readable code, but if it makes you confuse, then it totally fails its purpose and should be refactored, wdyt?

src/components/accessories/admin/users/usersTable/UsersTable.tsx
mwithi: Sorry, where is saved in plain text in the database? it should not... it should be encrypted and only encrypted passwords should be sent to the BE for comparing

mwithi: Or at least in this way is done using GUI that relies on the same DB tables

mwithi: Up!

src/components/accessories/admin/users/editUser/EditUser.tsx
SilverD3: Sorry @gasp , I was busy with something else and I completely forgot to check your PR from time to time.

You can still create another state, maybe `state.users.getByid` or `state.users.getByUsername`. That way, you could dispatch an action that fetch a single user using his id or username, store the result it the state you'll have created, so that, later you can query it. 
Consider the first problem I mentioned 
> If the endpoint to get all users is paginated, you might not find the user even though he exists

Because if there are more that a hundred user, I don't think we're going to fetch them all at once. If the endpoint we use to fetch users is not yet paginated, we should paginate it.

src/components/accessories/admin/users/editUser/EditUser.tsx
SilverD3: Well, let's just say we've been doing it ever since, and I don't think it's a problem. If you read a state slice just to pass it to a child component (which I consider as a kind of `props drilling`), I think you could better read the state directly within the child component.

src/components/accessories/admin/users/usersTable/UsersTable.tsx
SilverD3: Exactly @mwithi . As @SteveGT96 is working on OH2-253, I was drawing his attention to this point. Password, even encrypted, should not be sent to FE

mwithi: I get a "not allowed" error when changing a user's group, so I think we have to wait for OH2-253 to be merged

image

src/components/accessories/admin/users/editUser/EditUser.tsx
gasp: useFormik does not accept new initalProps, only the ones available on mount

src/components/accessories/admin/users/editUser/EditUser.tsx
gasp: 100 user payload is less than 100 kb
If we add pagination, we will have to rework entirely table filtering.

src/components/accessories/admin/users/editUser/EditUser.tsx
SilverD3: I'm not sure what this has to do with my previous comment.

src/components/accessories/admin/users/editUser/EditUser.tsx
SilverD3: I still believe that here, you have to get a single user, not all the user in the app.

SteveGT96: Hi @gasp ! Thanks for the efforts.
I don't think takink the user to update from the user list could be the best way(that could be error prone and not really easy to maintain).
That said, when the user click on the edit button then you navigate to /admin/users/:username/edit(or something like) and in the EditUser component, you load the the user matching the username param in the state(state.users.[user | getByUsername]).

Why that pattern ?

Modern application acts that way and considers it as a best practice.
- React router for example gives the possibility to fetch the user before rendering the target component(using data loader pattern, put in our case, we're already using redux).
- Angular for example has a built in approch using resolver guards.

SilverD3 (approved):

mwithi (approved):
LGTM. Thanks a lot!

PR #645, opened 2024-08-06
chore: fix unused imports

gasp:
before: ✖ 545 problems (0 errors, 545 warnings)
after: ✖ 270 problems (0 errors, 270 warnings)

As mentionned here https://github.com/informatici/openhospital-ui/pull/641#issuecomment-2268398013 there is a lot of unused imports in the codebase (and it has been for a long time)

I recommend using VSCode auto-organize import feature (it removes unused imports) https://code.visualstudio.com/docs/languages/javascript#_organize-imports

PR #648, opened 2024-08-09
chore: allow sharing common vscode config

gasp:
Howdy vscode users 👋🏼 , should we share common vscode config ?

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -22,7 +22,6 @@ storybook-static
 npm-debug.log*
 yarn-debug.log*
 yarn-error.log*
-.vscode
 
 # IntelliJ
 .idea


.gitignore
```diff
@@ -22,7 +22,6 @@ storybook-static
 npm-debug.log*
 yarn-debug.log*
 yarn-error.log*
-.vscode
```
mwithi, line 25: should .vscode settings need to be included (bad habit) or we can just use a different approach as I was suggesting here?

https://github.com/informatici/openhospital-ui/pull/645#issuecomment-2277401833

SilverD3: I think we can include `.vscode` file to let vscode users share useful configs. `.ide-settings/` is generic and is not automatically processed by vscode, I think.

mwithi: is not possible to add a `vscode` folder in `.ide-settings/` like we have done for other IDEs?

SilverD3: Yeah, we can do that. 
So @gasp if you agree, create the folder `.ide-settings`, add .vscode in that folder. So that you can let .vscode in the root folder excluded.

mwithi: so we can close this PR?

SilverD3: If @gasp agree, yes

PR #658, opened 2024-09-10
fix: strict equals instead of abstract equals

gasp:
(minor)

It is considered good practice to use the type-safe equality operators === and !== instead of their regular counterparts == and !=.

The reason for this is that == and != do type coercion which follows the rather obscure [Abstract Equality Comparison Algorithm](https://www.ecma-international.org/ecma-262/5.1/#sec-11.9.3). For instance, the following statements are all considered true:
* [] == false
* [] == ![]
* 3 == "03"

([source](https://eslint.org/docs/latest/rules/eqeqeq))


SilverD3 (approved):
Thanks @gasp . When we run eslint, we can see almost a hundred of this warning (`Expected '===' and instead saw '=='`)

mwithi (approved):
Reviewed and approved by @SilverD3

PR #677, opened 2024-10-17
OH2-404 | Implement supplier deletion

SilverD3:
See OH2-404


SteveGT96 (approved):

mwithi (approved):
Might be useful in the future to "see" the deleted records, similarly as we already do for disabled diseases. Out of this, LGTM. Thanks

PR #682, opened 2024-10-24
OH2-396 | Add a lock mechanism in all the remaining entities without it

SilverD3:
See OH2-396
Paired with [#479](https://github.com/informatici/openhospital-api/pull/479)


mwithi (approved):
LGTM. Thanks a lot!

PR #700, opened 2024-11-25
OP-1043 | Remove deleted group from user edit group jcombo box

SteveGT96:
See [OP-1043](https://openhospital.atlassian.net/browse/OP-1043)

[OP-1043]: https://openhospital.atlassian.net/browse/OP-1043?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/src/components/accessories/admin/users/editUser/EditUser.tsx
+++ b/src/components/accessories/admin/users/editUser/EditUser.tsx
@@ -65,7 +65,7 @@ export const EditUser = () => {
 
   return (
     <EditUserForm
-      groups={groups.data}
+      groups={groups.data?.filter((group) => !group.deleted)}
       initialValues={user}
       onSubmit={handleUpdate}
       isLoading={isLoading}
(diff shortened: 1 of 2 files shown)


mwithi:
Found that both User edit and Group edit do not have CANCEL option to go back without doing changes.
Moreover in User edit, in the column Groups, I wouls show the group name rather than its description, or both, by the moment that the column Deleted could be reduced in size.

Let me know if you want/could address these issues in this same PR or another issue (in which case we can merge this one).
Thanks

SteveGT96:
> Found that both User edit and Group edit do not have CANCEL option to go back without doing changes. Moreover in User edit, in the column Groups, I wouls show the group name rather than its description, or both, by the moment that the column Deleted could be reduced in size.
> 
> Let me know if you want/could address these issues in this same PR or another issue (in which case we can merge this one). Thanks

Let's address that in a separate issue. Thanks.

mwithi:
> Let's address that in a separate issue. Thanks.

Do you create the issue or I do?

SteveGT96:
Please, do it !

mwithi (approved):
LGTM. Thanks a lot!

PR #709, opened 2024-12-10
OH2-418 | Orthanc Integration

SilverD3:
See OH2-418


mwithi:
The feature will be developed in a dedicated branch

PR #712, opened 2025-01-21
OP-1369 Add files for webserver deployment

mwithi:
See OP-1369.

Hi guys, how do you feel with this addition to the public/ folder? Make sense to have a "build-ready" for application server deployment (e.g. tomcat)?


mizzioisf (approved):
I think this is good. We might move the file somewhere else in the future but I would approve and go ahead with the testing now.

SilverD3:
@mwithi this file should be add in the API, I think.

mwithi:
No longer needed

PR #715, opened 2025-01-28
Fix redux typescript

alefalezza:
Just fixed some warnings I got starting application locally with mocks.

### DIFF
--- a/src/components/accessories/dashboard/Dashboard.tsx
+++ b/src/components/accessories/dashboard/Dashboard.tsx
@@ -1,3 +1,4 @@
+import { createSelector } from "@reduxjs/toolkit";
 import { Chart, registerables } from "chart.js";
 import { useAppSelector } from "libraries/hooks/redux";
 import React from "react";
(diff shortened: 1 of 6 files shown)


src/components/accessories/dashboard/Dashboard.tsx
```diff
@@ -10,12 +11,15 @@ import "./styles.scss";
...
+  (userCredentials) => ({ userCredentials })
+);
+
 const Dashboard = () => {
   const { t } = useTranslation();
 
-  const { userCredentials } = useAppSelector((state) => ({
-    userCredentials: state.main.authentication.data,
-  }));
+  const { userCredentials } = useAppSelector(appSelector);
```
SilverD3, line 22: @alefalezza what is wrong with the actual implementation?

SilverD3 (commented):
What is the plus when you create a selector before using it? I mean, what issue are you fixing? Why not to do the same every time we getting a state slice?

src/components/accessories/dashboard/Dashboard.tsx
SteveGT96: @SilverD3, the actual implementation could cause performence issues. Every time an action is dispatched, the value is computed(even if it has not been changed). I think we should refactor the whole App while taking that into account.

SteveGT96: @alefalezza Nice job, I think once this merged, we'll make a refactoring and extract reusable selectors in state/[namespace]/selectors.ts.

SteveGT96 (approved):

mwithi (approved):
Reviewed and approved by @SteveGT96

PR #725, opened 2025-03-28
OH2-462 | Fix error on the message for canceling and deleting of laboratory exam

ArnaudFofou:
See OH2-462.


mwithi (approved):
LGTM. Thanks a lot!

PR #728, opened 2025-04-16
OH2-464 Switch to informatici GitHub pages

mwithi:
See OH2-464.

### DIFF
--- a/.github/workflows/deploy.yml
+++ b/.github/workflows/deploy.yml
@@ -55,6 +55,6 @@ jobs:
         run: npm run clear && npm run build:gh-pages
       - run: git add --force build
       - run: git commit -m 'new build'
-      - run: git remote add intesys-oh https://github.com/intesys/openhospital-ui.git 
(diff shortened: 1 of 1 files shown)


.github/workflows/deploy.yml
```diff
@@ -55,6 +55,6 @@ jobs:
         run: npm run clear && npm run build:gh-pages
       - run: git add --force build
       - run: git commit -m 'new build'
-      - run: git remote add intesys-oh https://github.com/intesys/openhospital-ui.git 
+      - run: git remote add oh-ui-pages https://github.com/informatici/openhospital-ui.git
```
SilverD3, line 59: Okay, but here we don't need to add another remote repo because `origin` is already pointing `https://github.com/informatici/openhospital-ui.git`. But we can keep it, in case we later change the remote repo for github pages

mwithi: Good point, I've put a comment as reminder, thanks!

PR #776, opened 2026-06-19
OP-1393 Add license-header tooling for the UI

giuseppesorge:
## OP-1393 — File license header template for UI

Gives the UI repo a license-header system like the Java modules have:
- **`.license-header.txt`** — GPL-3.0 header for `.ts`/`.tsx` files (adapted from the core module's template).
- **`license-check-and-add-config.json`** — targets `src/**/*.{ts,tsx}`, excludes generated/codegen (`src/generated`, `src/__generated__`, `*.d.ts`) and config files.
- **`package.json`** — declares `"license": "GPL-3.0"` and adds `license:check` / `license:add` npm scripts (+ `license-check-and-add` devDependency).

`npm run license:check` runs correctly and reports the source files missing the header (973 `.ts`/`.tsx` under `src/`).

**Scope note:** this PR introduces the tooling/template only. Actually **stamping** the header onto the existing ~973 files is a deliberate follow-up (a single `npm run license:add`), so this infrastructure change stays reviewable and the maintainer can confirm scope (e.g. whether to also cover `cypress/`, `scripts/`) before the mass edit.

### DIFF
--- a/.license-header.txt
+++ b/.license-header.txt
@@ -1,0 +1,21 @@
+/*
+ * Open Hospital (www.open-hospital.org)
+ * Copyright © 2006-2026 Informatici Senza Frontiere (info@informaticisenzafrontiere.org)
+ *
+ * Open Hospital is a free and open source software for healthcare data management.
+ *
+ * This program is free software: you can redistribute it and/or modify
+ * it under the terms of the GNU General Public License as published by
+ * the Free Software Foundation, either version 3 of the License, or
+ * (at your option) any later version.
+ *
+ * https://www.gnu.org/licenses/gpl-3.0-standalone.html
+ *
+ * This program is distributed in the hope that it will be useful,
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+ * GNU General Public License for more details.
+ *
+ * You should have received a copy of the GNU General Public License
+ * along with this program. If not, see <https://www.gnu.org/licenses/>.
+ */
--- a/license-check-and-add-config.json
+++ b/license-check-and-add-config.json
@@ -1,0 +1,44 @@
+{
+	"license": ".license-header.txt",
+	"licenseFormats": {
+		"ts|tsx": {
+			"file": ".license-header.txt"
+		}
+	},
+	"trailingWhitespace": "TRIM",
+	"ignoreFile": ".gitignore",
+	"ignore": [
+		"node_modules/**",
+		"build/**",
+		"dist/**",
+		"coverage/**",
+		"storybook-static/**",
+		"public/**",
+		"scripts/**",
+		"cypress/**",
+		"docs/**",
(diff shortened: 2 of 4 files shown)


duval2dev (commented):
@giuseppesorge — this looks good, two small things before it can be merged:
1. Could you rebase onto current `develop`? The branch is a couple months old and `package-lock.json` has drifted.
2. The header in `.license-header.txt` says "Copyright © 2006-2025" — might want to update that to 2026 (or a range through the current year) before it gets stamped onto ~974 files.

Once that's done, @mwithi can review and merge.

.license-header.txt
```diff
@@ -0,0 +1,21 @@
+/*
+ * Open Hospital (www.open-hospital.org)
+ * Copyright © 2006-2025 Informatici Senza Frontiere (info@informaticisenzafrontiere.org)
```
mwithi:
```suggestion
 * Copyright © 2006-2026 Informatici Senza Frontiere (info@informaticisenzafrontiere.org)
```

mwithi: this will be automatically updated every time we update one file?

duval2dev: Doesn't look like it. I don't see a year placeholder in .license-header.txt, and license-check-and-add seems to just do a literal text match, so I'd guess it stays static once stamped. Worth double-checking though.

giuseppesorge: Applied in 1533ae63 — the branch was rebased, so I committed the change rather than clicking the suggestion.

> this will be automatically updated every time we update one file?

No — I verified with the tool itself (v4.0.5) on scratch files: the header is matched and stamped **literally**, there is no year placeholder or templating. Concretely:

- once files are stamped, they keep whatever year the template had at stamping time;
- if `.license-header.txt` is later edited (say to 2006-2027), `license:check` reports **every** stamped file as missing the license, and running `license:add` at that point prepends a second header above the old one instead of updating it.

So a future year bump would be a one-shot find-and-replace across the stamped files (plus the template), not a `license:add` run. Updating the template to 2026 now, before the mass stamping, costs nothing — thanks for catching it before it landed on ~978 files.

While testing this I also noticed the previous `licenseFormats` shape (`"prepend": ""`/`"append": ""`) made `add` insert a blank line before the header. I switched it to `{"file": ".license-header.txt"}` in the same commit: stamping a copy of `src/index.tsx` now puts the header at the very top of the file and `check` passes on the result.

giuseppesorge:
@duval2dev thanks for the review — both points addressed:

1. Rebased onto current `develop`; `package-lock.json` was regenerated from `package.json` (`npm install --package-lock-only`) rather than hand-merged, same as on the other UI PRs. `license:check` runs against the rebased tree and reports 978 files missing the header (develop gained a few files since July), consistent with the tooling-only scope.
2. Header year updated to 2006-2026 along with @mwithi's suggestion — details on how the tool treats the year are in the inline thread.

duval2dev (approved):

mwithi (approved):

PR #783, opened 2026-07-16
OP-1430 Validate passwords against the backend policy in the web UI

giuseppesorge:
Part of OP-1430: https://openhospital.atlassian.net/browse/OP-1430

The web UI now fetches the password policy from `GET /auth/password-policy` into a redux `passwordPolicy` slice via a `usePasswordPolicy` hook, and validates passwords with a shared `passwordPolicySchema` yup builder — used by the change-password screen **and** the create/edit-user admin forms — replacing three hardcoded, drifted rule sets (`MIN_PASSWORD_LENGTH` / `STRONG_PASSWORD_REGEX` in change-password, and a different `passwordRules` min-5 + upper/lower/digit in the admin forms that no longer matched what the API has enforced since OP-896).

- The minimum length applies always (0 = none) and the strength regex only when the policy is enabled, mirroring the core `isPasswordValid`. Both checks tolerate an empty value, so required forms surface their own "required" message and the optional edit-user password (blank = no change) still saves.
- The now-inaccurate i18n messages that described the old rule are made generic.
- The endpoint is mocked in both the PollyJS and MSW layers (shared fixture) so the forced-change flow stays demoable without a backend.

Unit tests for `passwordPolicySchema`; verified live in mock mode (change-password and create-user reject a password the old rule accepted and accept a compliant one).

Companion PRs: core informatici/openhospital-core#1647 · api informatici/openhospital-api#607
Depends on the core + api PRs.


duval2dev:
@mwithi — recommend holding this one pending https://github.com/informatici/openhospital-core/pull/1647 (and https://github.com/informatici/openhospital-api/pull/607). 

Please could you mark it as `hold`?

mwithi:
> @mwithi — recommend holding this one pending [informatici/openhospital-core#1647](https://github.com/informatici/openhospital-core/pull/1647) (and [informatici/openhospital-api#607](https://github.com/informatici/openhospital-api/pull/607)).
> 
> Please could you mark it as `hold`?

Hi @duval2dev you can just approve if you think can be merged. The merge action will be done at the right moment.
Thanks!

duval2dev (approved):

PR #785, opened 2026-07-17
OH2-483 Apply the safe Dependabot dependency bumps and auto-fixable runtime transitives

giuseppesorge:
## What

Applies the safe dependency bumps flagged by Dependabot, plus the auto-fixable runtime transitives:

- `lodash` `^4.17.21` → `^4.18.1` (direct runtime)
- `react-router` / `react-router-dom` `^7.10.1` → `^7.15.0` (resolves 7.18.1)
- `overrides`: `lodash-es` → `4.18.1`, `postcss` → `8.5.19` (vulnerable pinned transitives)

`npm audit`: **48 → 42** vulnerabilities (−4 high, −2 moderate).

## Why

Part of the Dependabot vulnerability cleanup (OH2-482 epic). This ticket covers only the low-risk bumps; the remaining alerts are tracked separately: critical `jspdf` (OH2-484), DOMPurify XSS (OH2-485), dev/build tooling (OH2-486).

## Verification

- `tsc --noEmit` + `vite build` green
- unit tests (vitest): 26/26 green
- full Cypress e2e suite: all specs passed (345 passing, 14 pending, 0 failing)
- live navigation smoke on the bumped `react-router` (login redirect, all main routes, nested admin redirect, deep-link, history back)

## Notes

- Deferred with rationale: the flagged `immutable` instance comes via `sass` (dev-only → OH2-486); `yaml@1.10.2` is build-time via `cosmiconfig` (forcing `yaml@2` breaks it) — both out of scope here.

Jira: OH2-483

PR #747, opened 2025-12-18
OH2-470 (SD-30 | SD-31 | SD-32) | Migrate to react 19, vite, msw

SteveGT96:
See OH2-470.

This pull request primarily migrates the project environment configuration and related tooling from the `REACT_APP_` convention to the `VITE_` convention, aligning the codebase with Vite as the build tool instead of Create React App. It also updates configuration files, scripts, and documentation to reflect this migration, adjusts development and CI/CD ports, and introduces a new Biome configuration for linting and formatting. Additionally, it standardizes code style in Cypress tests and Storybook configs.

**React 19 migration**

**Vite Migration and Environment Configuration:**

- Migrated all environment variable files (e.g., `.env`, `.env.development`, `.env.prod`, etc.) from `REACT_APP_` prefixes to `VITE_` prefixes, and updated related variable names to match Vite conventions. [[1]](diffhunk://#diff-e9cbb0224c4a3d23a6019ba557e0cd568c1ad5e1582ff1e335fb7d99b7a1055dL1-R2) [[2]](diffhunk://#diff-da4c41d59c967338247b2f0ea6d845ab05171f942fd9521b96d3304202185db9L1-R3) [[3]](diffhunk://#diff-7dd3b767860ba2de80c6d4c050db40dedd73b85a2a54a6eb3b5b54fd3d15dc56L2-R2) [[4]](diffhunk://#diff-b26af0b14c436b4390c93b145fca85f16b5bac28359a6ecdd6e4ea0524081b33L1-R4) [[5]](diffhunk://#diff-53003b50d89f576ea396020048779f70e1f21c14f36ac067c52c269e041c50d9L2-R2) [[6]](diffhunk://#diff-bd8c5a5839b43575d62ee4f8b0cf6688f479ddd92251d43708b17259a48997baL1-R1)
- Updated documentation in `README.md` to reference new `VITE_` environment variables. [[1]](diffhunk://#diff-b335630551682c19a781afebcf4d07bf978fb1f8ac04c6bf87428ed5106870f5L34-R34) [[2]](diffhunk://#diff-b335630551682c19a781afebcf4d07bf978fb1f8ac04c6bf87428ed5106870f5L43-R44)

**Tooling and Configuration Updates:**

- Added a new `biome.json` configuration file to enable Biome for linting, formatting, and code assistance across the project.
[...]

PR #786, opened 2026-07-17
OH2-484 Replace react-component-export-image with direct html2canvas and jsPDF

giuseppesorge:
## What

Replaces `react-component-export-image` with direct `html2canvas` + `jspdf@4.2.1` in `DataDownloadButton` (the dashboard widgets' PDF/PNG export), keeping the same A3-landscape PDF layout and file naming. The widget refs are already DOM refs, so nothing else changes for the callers.

## Why

Two problems, one root cause — the wrapper is unmaintained (last release 2022):

1. **Security (this ticket):** it pins a `jspdf` affected by 12 advisories, up to critical PDF-injection ones, with no fix available through the wrapper (`fixAvailable: false`).
2. **The export is currently broken on develop:** the wrapper relies on `ReactDOM.findDOMNode`, which React 19 removed, so every PDF/PNG export crashes with `TypeError: findDOMNode is not a function` since the React 19 migration (#747). Reproduced on develop; no test covers this path, which is why it went unnoticed.

## Security impact

`npm audit` (on top of #785): **42 → 39** — clears the `jspdf` **critical**, the `react-component-export-image` **high**, and the `dompurify` advisory (jsPDF 4 bundles the patched DOMPurify), which de-facto resolves **OH2-485** as well.

## Verification

- `tsc` + `vite build`, biome lint, unit tests 26/26
- Live export test on the dev server: PDF generated and opened (A3 landscape, chart rendered correctly, 2.4 MB), PNG generated (verified blob + file download)
- The pre-existing crash was reproduced on a branch without these changes to confirm it is independent of this PR

## Notes

- **Stacked on #785 (OH2-483)** — please merge that one first; until then this diff shows both commits.
- Found while testing: the mock server's `GET /usersettings/dashboard` returns `{}` (TODO) and there is no save route, so the dashboard renders no widgets in mock mode and shows "can't save customization" — worth a small follow-up on the mocks (an unused `dashboardSettingDTO` fixture already exists).

Jira: OH2-484


giuseppesorge:
Rebased onto current `develop` and lockfile regenerated from `package.json` — this was conflicting on the package files against the Dependabot bumps that landed since. `tsc`, `vite build` and `vitest` all green. Details and the audit numbers for the whole stack are in #788.

duval2dev (approved):

duval2dev (approved):
@mwithi — #785 and #786 are both reviewed and good to merge. Please merge #785, then #786, then #788 directly after — it's a stack, so #788's diff will shrink to just its own dev-tooling scope once the first two are in.

mwithi (approved):

PR #787, opened 2026-07-17
OH2-475 Improve mobile usability: header menu navigation and mobile scroll/layout

giuseppesorge:
### Navigation (mobile header menu)
- The mobile menu now closes and restores body scrolling on **every route change** (nav items and the logo/breadcrumb links), via a `useLocation` effect.
- Tapping the nav item of the **current route** also closes the menu — same-route navigation does not change the pathname, so the menu is closed explicitly (`navigateFromMenu`).
- `openMenu` toggles the `disable-scroll` body class cleanly, so the scroll lock is never left orphaned.

### Scroll / layout
- Full-viewport activity containers use `min-height` instead of a fixed `height: calc(100vh - …)`, so pages scroll naturally on phone viewports.
- The patient-search container is taken out of `position: absolute` (now `position: relative` + `min-height`) so it can scroll to its footer.
- The dashboard period picker wraps within phone widths instead of overflowing.

### Dashboard (phone breakpoint)
- `getBreakpointFromWidth` fell back to `md` below the `xs` breakpoint, so phone widths rendered the two-column `md` layout and squeezed the widgets to half-width. Sub-`xs` widths now map to `xxs` (full-width, stacked), and drag/resize/drop are disabled on that breakpoint (awkward on touch).
- The breakpoint helper and its config were extracted into a window-free module and covered by a unit test.

### Shared modal
- The shared `CustomModal` used fixed per-breakpoint widths (800/600/400/300px). Replaced with a single responsive rule — up to 800px, capped to the viewport minus a margin — so it fits on any width and uses the available space on phones.

### Tests
- `cypress/integrations/mobile_usability.cy.ts`: 5 specs at a 390×844 viewport (menu-close-on-navigation via nav item / current route / logo link, dashboard document flow, patient-search scroll). All green (run with the `--lang=en-US` electron flag).
- `breakpoints.test.ts`: unit test for the phone-breakpoint mapping.

Remaining for a follow-up: responsive max-width rules on the other shared fixed-width fields/buttons/panels (a live overflow audit across pages).

### DIFF
--- a/cypress/integrations/mobile_usability.cy.ts
+++ b/cypress/integrations/mobile_usability.cy.ts
@@ -1,0 +1,70 @@
+/// <reference types="cypress" />
+
+describe('Mobile usability spec', () => {
+	beforeEach(() => {
+		cy.viewport(390, 844);
+	});
+
+	it('should close the mobile menu and restore scrolling when navigating', () => {
+		cy.authenticate('/dashboard');
+
+		cy.dataCy('app-header-identified-trigger').click();
+		cy.dataCy('app-header').should('have.class', 'open_menu');
+		cy.get('body').should('have.class', 'disable-scroll');
+
+		cy.get('.appHeader__nav__item').contains(/patients/i).click();
+
+		cy.url().should('include', '/patients');
+		cy.dataCy('app-header').should('not.have.class', 'open_menu');
+		cy.get('body').should('not.have.class', 'disable-scroll');
+	});
+
+	it('should close the mobile menu when tapping the nav item of the current route', () => {
+		cy.authenticate('/patients');
+
+		cy.dataCy('app-header-identified-trigger').click();
+		cy.dataCy('app-header').should('have.class', 'open_menu');
+		cy.get('body').should('have.class', 'disable-scroll');
+
+		// same-route navigation: the pathname does not change, the menu must close anyway
+		cy.get('.appHeader__nav__item').contains(/patients/i).click();
+
+		cy.url().should('include', '/patients');
+		cy.dataCy('app-header').should('not.have.class', 'open_menu');
+		cy.get('body').should('not.have.class', 'disable-scroll');
+	});
+
+	it('should close the mobile menu when navigating through the logo link', () => {
+		cy.authenticate('/patients');
+
+		cy.dataCy('app-header-identified-trigger').click();
+		cy.get('body').should('have.class', 'disable-scroll');
+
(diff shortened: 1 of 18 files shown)


duval2dev:
@giuseppesorge — manual QA on this one turned up one real issue: on a phone-width first load (hard reload directly at e.g. 390px), there's a visible flash where drag/resize briefly look enabled before snapping off.

Root cause: `localBreakpoint` in `GridLayoutContainer.tsx` defaults to `'md'` (`useState<string>('md')`) until `onBreakpointChange` fires, so `isTouchLayout` is `false` for that first frame even though the grid is already rendering at `xxs`. Since `getBreakpointFromWidth` is already imported in this file, could you switch the initializer to compute it from `window.innerWidth` up front, e.g.:
```ts
const [localBreakpoint, setLocalBreakpoint] = useState<string>(() =>
  typeof window !== 'undefined' ? getBreakpointFromWidth(window.innerWidth) : 'md',
);
```
Everything else looks good:  nav menu, scroll/layout, and modal responsiveness all checked out on manual QA. Once this is in, good to merge.

giuseppesorge:
Good catch, thank you — and thanks for the QA pass. Fixed in fc29704b, with your initializer: the state now starts from `getBreakpointFromWidth(window.innerWidth)`, so the first frame already knows it is on a phone and drag/resize are never enabled.

On verification, so you know exactly what I did and did not check: `getBreakpointFromWidth` is already unit-tested and maps 390 to `xxs`, so the initial value is the same one the effect would have set a frame later; typecheck and biome are clean, and `mobile_usability.cy.ts` still passes 5/5. I could not reproduce the flash in cypress — with the mocked API the dashboard renders without any widgets, so `.react-grid-layout` never mounts and there is nothing to observe. If you have a moment, a second look at phone-width first load on your side would be the real confirmation.

giuseppesorge:
Also rebased onto current `develop` — the branch was 15 commits behind, so it still carried `react-component-export-image` that OH2-484 has since removed. The rebase was clean: none of the 9 commits here touch a file develop changed.

Two things worth a note, since the force-push moved the commits:
* the breakpoint fix is now `8e10c960`, same content as before;
* one extra commit runs the project formatter over the two files this branch touches (an import list in `AppHeader.tsx` and a cypress chain), which the formatter wanted split across lines. Nothing else changed.

CI is green on the rebased branch — build, lint, type-check, unit tests and e2e.

duval2dev (approved):

mwithi (approved):

PR #804, opened 2026-08-25
chore(deps): bump vite from 7.2.7 to 8.2.2 (+ cypress-vite/chokidar fix)

duval2dev:
See [PR-#798](https://github.com/informatici/openhospital-ui/pull/798)


bun.lock
mwithi, line 1: bun.lock has been commited recently (24d0eb5083b23cc1f0f231cbbd210c9bcc21f90e) maybe by mistake, I think we can remove it

PR #809, opened 2026-08-28
OH2-488 OH2-489 Fix laboratory date input validation

mwithi:
See OH2-488 OH2-489.

Fixes date handling in the Laboratory filter:
- preserves partially typed dates without resetting or shifting the selected day;
- prevents malformed or intermediate values from crashing date conversion;
- uses local calendar dates instead of UTC conversion;
- updates date values and related Formik state atomically;
- prevents stale cross-field validation errors between Date From and Date To;
- keeps the draft-input behavior scoped to the Laboratory date range fields;
- changes the default Laboratory date range to yesterday–today to avoid retrieving unnecessary historical records.

### DIFF
--- a/src/components/accessories/dateField/DateField.tsx
+++ b/src/components/accessories/dateField/DateField.tsx
@@ -8,6 +8,7 @@ import type { IProps } from './types';
 const DateField: FunctionComponent<IProps> = ({
 	fieldName,
 	fieldValue,
+	preserveDraftInput = false,
 	disableFuture,
 	disabled,
 	label,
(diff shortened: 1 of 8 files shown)


duval2dev (approved):
@mwithi — one quick check: `consts.ts` also changes the default filter date range (from the old, inconsistent `startOf('month')`/`subtract(1, 'years')` split to a consistent `subtract(1, 'day')` for both). Wasn't in the PR description's bullet.
Just confirming that's intentional as part of the fix?

mwithi:
> @mwithi — one quick check: `consts.ts` also changes the default filter date range (from the old, inconsistent `startOf('month')`/`subtract(1, 'years')` split to a consistent `subtract(1, 'day')` for both). Wasn't in the PR description's bullet. Just confirming that's intentional as part of the fix?

Good catch! And yes, that change is intentional. The default Laboratory filter range was changed to yesterday–today because laboratories typically work on one or two days of data, and loading a month or a year by default retrieves unnecessary records.
It also removes the previous inconsistency between the form’s initial fields (startOf('month')) and the submitted filter (subtract(1, 'years')).

Let me update the PR description.

PR #824, opened 2026-09-08
chore: Fix Sass and browserslist console warnings on startup

duval2dev:
@mwithi 
@almeida1492 

Startup (`npm start`) was printing ~230+ warnings in the console/terminal: a stale browserslist database, Sass deprecation warnings originating entirely from the (unmaintained) Susy grid library in `node_modules`, deprecated `lighten()`/`darken()` color functions in our own stylesheets, and 116 files still using the legacy `@import` syntax (removed in Dart Sass 3.0). This PR eliminates all of them.

### Changes

- **Browserslist**: updated the `caniuse-lite` database (`package-lock.json`) via `update-browserslist-db`.
- **`vite.config.ts`**: added `css.preprocessorOptions.scss.quietDeps: true` and `loadPaths: [node_modules]` so Dart Sass correctly classifies Susy as a dependency and silences warnings originating from its internal (legacy) code, instead of us patching a third-party package.
- **Deprecated color functions**: replaced all `lighten()`/`darken()` calls with `color.adjust($color, $lightness: ±N%)` (`sass:color` module) across 9 files (dashboard, patient details, search patient, laboratory, tables, opds, billing).
- **Full `@import` → `@use`/`@forward` migration**: converted every remaining `@import` statement (116 files) to `@use "..." as *;`, preserving identical bare (unprefixed) variable/mixin access so no call sites needed renaming. Added `@forward "./variables";` in `src/styles/index.scss` to keep transitive variable access working for the 2 files that import `styles/index` rather than `styles/variables` directly.

### Result

- Zero console warnings on `npm start`.
- `npm run build`, `npm run lint`, and `npm run test` all pass cleanly.
- No visual/behavioral changes intended — this is a syntax/tooling modernization only (`color.adjust()` is a math-equivalent replacement for `lighten()`/`darken()`, and `@use ... as *` preserves the exact same global-scope member access as the old `@import`).

### Proposal: remove the Susy dependency

Susy (`node_modules/susy`) is unmaintained since ~2016 and is only used in this codebase for its `susy-media($breakpoint)` mixin — actual layout throughout the app is already flexbox-based. Its own internal Sass files use deeply legacy syntax, which is why `quietDeps` was needed above just to keep the console clean.

I'd suggest a follow-up PR to drop Susy entirely and replace `susy-media()` with a small custom breakpoint mixin using the existing `$smartphone`/`$tablet_port`/`$tablet_land`/... variables in `_variables.scss`. This would remove an unmaintained dependency and let us stop relying on `quietDeps`/`loadPaths` as a workaround. Happy to scope and open that separately if there's agreement.


mwithi (approved):
LGTM. Thanks a lot!

